PR #1, opened 2018-06-25
Add custom btc method for getBlockByNumber and getBlockByHash

matthewjablack:
This PR adds custom methods to initial chain abstraction library. Specifically adding the equivalent to `eth_getBlockByNumber` and `eth_getBlockByHash` on the bitcoin side, by enabling `bitcoin.getBlockByNumber` and `bitcoin.getBlockByHash`. 

This is implemented by adding `caller` type to abstraction library, enabling calling of custom methods within bitcoin and ethereum drivers. 

```        
BTC                                                 ETH

Block Data
getblock "hash"                                     eth_getBlockByNumber int, bool, eth_getBlockByHash, "hash", bool
BTC data in common with ETH                         ETH block data in common with BTC
height:                                             number: => height
hash:                                               hash:
time: => timestamp                                  timestamp: => hex - int
difficulty:                                         difficulty:
size:                                               size:
previousblockhash: => parentHash                    parentHash:
nonce:                                              nonce:
BTC block data not in common with ETH               ETH block data not in common with BTC
confirmations:                                      mixHash:
strippedsize:                                       sha3Uncles:
weight:                                             logsBloom:
^ersion:                                            transactionsRoot:
versionHex:                                         stateRoot:
merkleroot:                                         receiptsRoot:
mediantime:                                         miner:
bits:                                               totalDifficulty:
chainwork:                                          extraData:
nextblockhash                                       gasLimit:
                                                    gasUsed:

http://chainquery.com/bitcoin-api/getblock          https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getblockbyhash


Transaction Data
gettransaction "txid"                               eth_getTransactionByHash "hash"
BTC transaction data in common with ETH             ETH transaction data in common with BTC
txid:  => hash:                                     hash:
amount: => value:                                   value:
blockhash: => blockHash                             blockHash:
blockindex:  => blockNumber                         blockNumber:
BTC transaction data in common with ETH             ETH transaction data not in common with BTC
blocktime:                                          nonce:
confirmations:                                      transactionIndex:
generated:                                          from:
walletconflicts:                                    to:
time:                                               gas:
timereceived:                                       gasPrice:
bip125-replaceable':                                input:
details:
hex:
details:
    account:
    address:
    category:
    amount:
    label:
    vout:
    fee:
    abandoned:

http://chainquery.com/bitcoin-api/gettransaction    https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionbyhash
```


dist/caller.js
```diff
@@ -0,0 +1,40 @@
+"use strict";
```
monokh, line 1: Should we `gitignore` the dist folder and remove it from the repo? Would make merge requests easier to follow.

src/drivers/ethereum/methods.js
```diff
@@ -37,5 +38,21 @@ export default {
         return new BigNumber(number.toString(10), 10)
       }
     }
+  },
+  eth_getBlockByNumber: {
+    formatter: {
+      output (object) {
+        object = renameKey(object, 'number', 'height')
```
monokh, line 45: I see that we have chosen `height` as the model's correct way of referring to the number of blocks. Have we made a conscious decision there on which is more commonly mentioned? 
If you have figured out the data model requirements, would be good if we could have a place to refer to it.

src/drivers/util.js
```diff
@@ -0,0 +1,11 @@
+export const asyncForEach = async (array, callback) => {
+  for (let index = 0; index < array.length; index++) {
+    await callback(array[index], index, array)
+  }
+}
+
+export const renameKey = (object, oldKey, newKey) => {
```
monokh, line 7: Consider here that you are actually changing the original object at the same time as returning it. 
It's best if you keep objects immutable while doing this. 
Use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

src/drivers/bitcoin/methods.js
```diff
@@ -198,5 +209,45 @@ export default {
     version: '>=0.1.0'
   },
   getAddressBalance: { version: '>=0.1.0' },
-  getAddressUtxos: { version: '>=0.1.0' }
+  getAddressUtxos: { version: '>=0.1.0' },
+  getBlockByNumber: {
+    version: '>=0.6.0',
+    custom: true,
+    function: {
+      async run (_this, params) {
```
monokh, line 228: I think i would prefer that the logic for these functions to be written in the module and referenced rather than inline to this big config. This way config can still be read easily when the methods have a lot of logic

monokh:
Did we agree on the general architecture for how integration is done with the chains? I know @harshjv  had some ideas here

harshjv:
Thanks, @mattBlackDesign for the PR. @monokh yes, I am currently modifying the architecture and it is a WIP. You can meanwhile, push changes to `dev` branch. We can reuse snippets from old branch later.

dist/caller.js
matthewjablack: Absolutely. Created PR to address this: https://github.com/ConsenSys/chainabstractionlayer/pull/2

src/drivers/ethereum/methods.js
matthewjablack: Added reference in PR description. Based on our discussion, it sounds like `height` is appropriate, but we may want to include multiple keys to ensure we are not enforcing a standard on developers.

src/drivers/bitcoin/methods.js
matthewjablack: Agreed. Sounds like this will be evident with the implementation of providers that @harshjv is adding. Will wait for that before merging or finalizing this.

dist/caller.js
harshjv: Because we are using ES2017 and later, we need to transpile the code and keep it in `dist` folder.

harshjv: without transpilation, the code will break.

monokh: @harshjv transpiling will still be there but I don't see why it needs to be included in the repo if the transpiled code is not being referenced externally. Anywhere we need the transpiled version, we can transpile it there

harshjv: Do `npm install react` and look into `node_modules/react`. See npm logs too.

matthewjablack:
Unnecessary with new newarch branch: https://github.com/ConsenSys/chainabstractionlayer/tree/newarch

PR #7, opened 2018-07-10
Add block schema in proper format

matthewjablack:


schema/block.json
```diff
@@ -0,0 +1,201 @@
...
+      "type": "string",
+      "title": "Previous Block Hash",
+      "pattern": "^0x([A-Fa-f0-9]{64})$"
+    },
+    "nonce": {
+      "type": "number",
+      "title": "Nonce",
+      "minimum": 0
+    },
+    "_raw": {
```
monokh: Thought we agreed defining the properties in `_raw` wasn't really necessary? @mattBlackDesign @harshjv

schema/block.json
```diff
@@ -0,0 +1,201 @@
+{
+  "$id": "https://dev.liquality.com/schema/block.json",
+  "title": "Block",
+  "description": "Blockchain block",
+  "type": "object",
+  "required": [ "number", "hash", "timestamp", "difficulty", "size", "previousBlockHash", "nonce" ],
```
monokh: what would `nonce` be in bitcoin?

monokh: are these limited to chain commonalities? notice `size` in there too

schema/block.json
```diff
@@ -0,0 +1,201 @@
...
+        "gasLimit": {
+          "type": "number",
+          "title": "Gas Limit",
+          "minimum": 0
+        },
+        "gasUsed": {
+          "type": "number",
+          "title": "Gas Used",
+          "minimum": 0
+        },
```
harshjv: SyntaxError! 🖌

schema/block.json
harshjv: @mattBlackDesign according to yesterday's discussion, we are just focusing on commonalities. You can safely remove `properties` of `_raw`. just make it `object`.

matthewjablack: Updated

schema/block.json
matthewjablack: `nonce` in btc is used for POW in the process of mining

Yes, these are chain commonalities

schema/block.json
```diff
@@ -0,0 +1,161 @@
...
+      "type": "string",
+      "title": "Parent Hash",
+      "pattern": "^0x([A-Fa-f0-9]{64})$"
+    },
+    "previousBlockHash": {
+      "type": "string",
+      "title": "Previous Block Hash",
+      "pattern": "^0x([A-Fa-f0-9]{64})$"
+    },
+    "nextBlockHash": {
```
harshjv: If `parentHash` is here, then what is the use of `previousBlockHash`?

schema/block.json
```diff
@@ -0,0 +1,161 @@
...
+      "type": "number",
+      "title": "Total Difficulty",
+      "minimum": 0
+    },
+    "extraData": {
+      "type": "string",
+      "title": "Extra Data",
+      "pattern": "^0x([A-Fa-f0-9]{64})$"
+    },
+    "gasLimit": {
```
monokh: the schema should only have properties defined for chain commonalities 
When the data is constructed, everything else can go in the `_raw` property - which doesn't need defining what properties it contains. its just a passthrough of all the chain specific properties that the chain may have.

matthewjablack: Agreed, added in following format:
```
"_raw": {
  "type": "object"
}
```

schema/block.json
matthewjablack: Updated to use `parentHash`

matthewjablack:
Updated @harshjv. Would love to get re-review before merging.

harshjv:
@mattBlackDesign can you move this file to `src/schema`? (PS. do rebase this before you push :) )

harshjv:
@mattBlackDesign did you rebase it?

monokh:
Github shows no changes? Is this branch and master really in sync? 😄

PR #10, opened 2018-07-17
Expose schemas on the client

monokh:
exporting so it's easy for other libraries (first the UI) to use it


harshjv:
I've pushed a commit to fix a style error. (Unexpected trailing comma, Newline required at end of file but not found.)

PR #13, opened 2018-07-25
Implement some Bitcoin RPC Provider Methods

i3wgnit:
Implemented the following:
- getBlockByHash
- getBlockByNumber
- getTransactionByHash
- getRawTransactionByHash
- _decodeRawTransaction


src/index.js
```diff
@@ -89,10 +95,36 @@ export default class Client {
   }
 
   /**
-   * Get block by number
-   * @param {!number} blockNumber - Number of the block to be fetched
-   * @param {boolean} [includeTx=false] - If true, fetches transaction in the block
-   * @return {Client.schemas.Block} Returns a Block
+   * Get a block given its hash.
+   * @param {!string} blockHash - A hexadecimal string that represents the *hash* of the desired block. If txHash is not a string, it will get converted through the `toString(16)` method.
```
monokh: I think the value of block hash should be consistent. So only accepts string and the user of the library can provide it like that

src/index.js
```diff
@@ -142,6 +183,27 @@ export default class Client {
     return transaction
   }
 
+  /**
+   * Get a raw hexadecimal transaction given its hash.
+   * @param {!string} txHash - A hexadecimal string that represents the *hash* of the desired transaction. If `hash` is not a string, it will be converted through the `toString(16)` method.
```
monokh: Similar to previous comment about consistent type for this parameter

monokh (commented):
Looks good. 

Good on removing the dist folder, it's very confusing when looking at PRs. Anyone that needs the dist can generate on demand with the code. 
Probably same for doc html, should figure a way to remove it since it's already inside the code.

monokh:
@harshjv the build somehow still failing on the node-hid installation 😿

harshjv:
@i3wgnit thanks for removing `dist` from git! I am also going to gitignore `docs`.

@mattBlackDesign any update on adding `libusb` to the docker image?

src/index.js
```diff
@@ -126,10 +162,15 @@ export default class Client {
     return blockHeight
   }
 
+  /**
+   * Get a transaction given its hash.
+   * @param {!string} txHash - A hexadecimal string that represents the *hash* of the desired transaction. If txHash is not a string, it will get converted through the `toString(16)` method.
```
harshjv: We are already expecting a string here. No need of `If txHash is not a string, it will get converted through the `toString(16)` method.`

src/providers/bitcoin/BitcoinRPCProvider.js
```diff
@@ -25,7 +25,76 @@ export default class BitcoinRPCProvider {
...
+      transactions,
+      confirmations }
+  }
+
+  async getBlockByNumber (blockNumber, includeTx) {
+    return this.getBlockByHash(await this._rpc('getblockhash', blockNumber), includeTx)
+  }
+
+  async getTransactionByHash (transactionHash) {
+    transactionHash = typeof transactionHash === 'string' ? transactionHash : transactionHash.toString(16)
```
harshjv: same here. `transactionHash` is a string. Add a check using `isString` instead.

src/schema/Block.json
```diff
@@ -21,7 +21,7 @@
       "minimum": 0
     },
     "difficulty": {
-      "type": "number",
+      "type": "string",
```
harshjv, line 24: Why `string`? Returning number in hex?

src/providers/bitcoin/BitcoinRPCProvider.js
```diff
@@ -25,7 +25,76 @@ export default class BitcoinRPCProvider {
...
+
+    const rawTx = await this.getRawTransactionByHash(transactionHash)
+    const tx = await this._decodeRawTransaction(rawTx)
+    const data = await this._rpc('gettransaction', transactionHash)
+
+    const { confirmations } = data
+    Object.assign(tx, { confirmations })
```
harshjv: Better way of using `Object.assign` is using traditional assignment statement and an empty object:

    const newObject = Object.assign({}, tx, { confirmations })

This also keeps original object `tx` intact.

harshjv (changes requested):
Do add `docs/` to .gitignore for now. I will setup docs building through CI later.

src/schema/Block.json
i3wgnit: `difficulty` is a number in scientific notation given as a string.

harshjv: Returning `string(number)` i.e. `"5178671069072.25"` makes more sense to me instead of scientific notation. Later, library like `BigNumber` can be used to manipulate the property.

src/schema/Block.json
```diff
@@ -21,7 +21,7 @@
...
       "title": "Difficulty",
       "minimum": 0
```
harshjv: String doesn't have `minimum` property in JSON Schema. You can use `minLength/maxLength` instead.

src/index.js
```diff
@@ -89,10 +95,40 @@ export default class Client {
...
 
   /**
-   * Get block by number
-   * @param {!number} blockNumber - Number of the block to be fetched
-   * @param {boolean} [includeTx=false] - If true, fetches transaction in the block
-   * @return {Client.schemas.Block} Returns a Block
+   * Get a block given its hash.
+   * @param {!string} blockHash - A hexadecimal string that represents the *hash* of the desired block.
+   * @param {boolean} [includeTx=false] - If true, fetches transaction in the block.
+   * @return {Promise<Client.schemas.Block, null>} Returns a Block with the same hash as the given input. Returns null if no block was found. If `includeTx` is true, the transaction property is an array of Transactions; otherwise, it is a list of transaction hashes.
```
harshjv: In `Promise<Client.schemas.Block, null>`, second parameter is for the error when `Promise` is rejected. So in this case, it will be `Error`

src/index.js
```diff
@@ -89,10 +95,40 @@ export default class Client {
...
+   */
+  async getBlockByHash (blockHash, includeTx = false) {
+    this._checkMethod('getBlockByHash')
+
+    if (!isString(blockHash)) {
+      throw new Error('Block hash should be a string')
+    }
+
+    if (!(/^(0x)?[A-Fa-f0-9]+$/.test(blockHash))) {
```
harshjv: Didn't we decide to enforce no `0x` prefix?

src/index.js
```diff
@@ -89,10 +95,40 @@ export default class Client {
...
+    }
+
+    return block
+  }
+
+  /**
+   * Get a block given its number.
+   * @param {!number} blockNumber - The number of the desired block.
+   * @param {boolean} [includeTx=false] - If true, fetches transaction in the block.
+   * @return {Promise<Client.schemas.Block, null>} Returns a Block with the same number as the given input. Returns null if no block was found. If `includeTx` is true, the transaction property is an array of Transactions; otherwise, it is a list of transaction hashes.
```
harshjv: Same here regarding the second parameter.

src/index.js
```diff
@@ -114,6 +150,10 @@ export default class Client {
     return block
   }
 
+  /**
+   * Get the current block height of the chain.
+   * @return {Promise<number>}
```
harshjv: `Error` as second parameter.

src/index.js
```diff
@@ -126,14 +166,48 @@ export default class Client {
     return blockHeight
   }
 
+  /**
+   * Get a transaction given its hash.
+   * @param {!string} txHash - A hexadecimal string that represents the *hash* of the desired transaction.
+   * @return {Promise<Client.schemas.Transaction, null>} Returns a Transaction with the same hash as the given input. Returns null if no transaction was found.
```
harshjv: Same here.

src/index.js
```diff
@@ -126,14 +166,48 @@ export default class Client {
...
+   */
   async getTransactionByHash (txHash) {
     this._checkMethod('getTransactionByHash')
 
-    if (!(/^[A-Fa-f0-9]$/.test(txHash))) {
```
harshjv, line 132: Can we check for hex here?

src/index.js
```diff
@@ -126,14 +166,48 @@ export default class Client {
...
+
+    return transaction
+  }
+
+  /**
+   * Get a raw hexadecimal transaction given its hash.
+   * @param {!string} txHash - A hexadecimal string that represents the *hash* of the desired transaction.
+   * @return {Promise<string, null>} Returns the raw Transaction with the same hash as the given output. Returns null if no transaction was found.
+   */
+  async getRawTransactionByHash (txHash) {
```
harshjv: Same here.

src/index.js
```diff
@@ -126,14 +166,48 @@ export default class Client {
...
+    if (!isString(txHash)) {
+      throw new Error('Transaction hash should be a string')
+    }
+
+    if (!(/^(0x)?[A-Fa-f0-9]+$/.test(txHash))) {
       throw new Error('Transaction hash should be a valid hex string')
     }
 
-    const transaction = await this.provider.getTransactionByHash(txHash)
+    const transaction = await this.provider.getRawTransactionByHash(txHash.replace('0x', ''))
```
harshjv: No need to replace `0x` if early checks throw an error on such strings.

harshjv (changes requested):
Looking forward to the updates!

src/index.js
i3wgnit: i have it a bit more down right?

harshjv (approved):
LGTM

PR #15, opened 2018-07-26
Implement generate swap for Bitcoin and Ethereum

i3wgnit:


src/index.js
```diff
@@ -247,6 +247,45 @@ export default class ChainAbstractionLayer {
...
+
+    if (!(/^[A-Fa-f0-9]+$/.test(secretHash))) {
+      throw new Error('Secret hash should be a valid hex string')
+    }
+
+    if (!isNumber(expiration)) {
+      throw new Error('Invalid expiration time')
+    }
+
+    const swap = this.provider.generateSwap(recipientAddress, refundAddress, secretHash, expiration)
```
monokh: can do return right here

src/providers/bitcoin/BitcoinLedgerProvider.js
```diff
@@ -36,4 +36,36 @@ export default class BitcoinLedgerProvider extends BitcoinRPCProvider {
...
+    //         # Check PKH and sig
+    // [77]    OP_EQUALVERIFY OP_CHECKSIG
+
+    expiration = expiration.toString(16)
+    let exLen = expiration.length
+    exLen += exLen % 2
+    expiration.padStart(exLen, '0')
+    exLen /= 2
+
+    return `76a97263a914${secretHash}8814${recipientAddress}67${exLen.toString(16).padStart(2, '0')}${expiration}b16d14${refundAddress}6888ac`
```
monokh: can this become multi line so it's a little easier to read

src/providers/bitcoin/BitcoinLedgerProvider.js
```diff
@@ -36,4 +36,36 @@ export default class BitcoinLedgerProvider extends BitcoinRPCProvider {
...
+    // [49+ex]     (?)(0x??) expiration[?]
+    // [50+ex]     OP_CHECKLOCKTIMEVERIFY
+    // [51+ex]     OP_2DROP
+    //             # Prep PKH check
+    // [72+ex]     (20)(0x14) refundPKH[20]
+    // [73+ex]  OP_ENDIF
+    //         # Check PKH and sig
+    // [75+ex] OP_EQUALVERIFY OP_CHECKSIG
+
+    expiration = expiration.toString(16)
```
monokh: it's generally not good to reassign variables
how about something like `const expirationHex = expiration.toString(16)`?

monokh (dismissed):
nice one, could you also bring the tests in here?

src/providers/bitcoin/BitcoinLedgerProvider.js
```diff
@@ -36,4 +36,38 @@ export default class BitcoinLedgerProvider extends BitcoinRPCProvider {
 
     return this._ledgerBtc.signMessageNew(this._derivationPath, hex)
   }
+
+  generateSwap (recipientAddress, refundAddress, secretHash, expiration) {
+    //         # Prep inputed PK for PKH check
```
harshjv: Should we move logic to generate swap in a separate swap-specific library?

i3wgnit: But its a pain to add new providers right now.
I can make a BitcoinSwapProvider.js that extends Ledger if you prefer

harshjv: I meant moving swap-specific logic and code to a new `liquality-swap` library maybe?

src/providers/ethereum/EthereumMetaMaskProvider.js
```diff
@@ -51,4 +51,24 @@ export default class EthereumMetaMaskProvider {
...
+    const expirationHex = expiration.toString(16)
+    const expirationEncoded = expirationHex.length % 2 ? '0' + expirationHex : expirationHex // Pad with 0
+    const expirationSize = expirationEncoded.length / 2
+    const redeemDestinationEncoded = (redeemDestinationBase + expirationSize).toString(16)
+    const refundDestinationEncoded = (refundDestinationBase + expirationSize).toString(16)
+    const expirationPushOpcode = (0x5f + expirationSize).toString(16)
+    const dataSizeEncoded = (dataSizeBase + expirationSize).toString(16)
+    const recipientAddressEncoded = recipientAddress.replace('0x', '') // Remove 0x if exists
+    const refundAddressEncoded = refundAddress.replace('0x', '') // Remove 0x if exists
+    const secretHashEncoded = secretHash.replace('0x', '') // Remove 0x if exists
```
matthewjablack, line 83: Some documentation comments would be nice for constants in generateSwap

src/index.js
```diff
@@ -247,6 +247,43 @@ export default class ChainAbstractionLayer {
...
+
+    if (!isString(refundAddress)) {
+      throw new Error('Refund address should be a string')
+    }
+
+    if (!isString(secretHash)) {
+      throw new Error('Secret hash should be a string')
+    }
+
+    if (!(/^[A-Fa-f0-9]+$/.test(recipientAddress))) {
```
monokh, line 333: probably better if we make this an `isHex` utility? then easier to read the validations because there are a lot 😄

src/providers/bitcoin/BitcoinLedgerProvider.js
monokh: I think it's fine to have the swap generation here, this way it's an interface of the CAL. If others want to develop swap generations that are different or for other chains, they can implement it as part of a provider. 
As a slightly neater solution, another provider could implement the generateSwap e.g. `BitcoinSwapProvider`? 
For the sake of getting stuff done, i think we should take the path of least resistance in getting the MVP working

src/providers/ethereum/EthereumMetaMaskProvider.js
monokh: This was my code actually 😄 
Perhaps add a little bit about the op codes or what the destinations mean? Generally, I would say it's pretty well documented just with how I split up every calculations name

src/providers/bitcoin/BitcoinLedgerProvider.js
```diff
@@ -36,4 +36,38 @@ export default class BitcoinLedgerProvider extends BitcoinRPCProvider {
...
+
+    let expirationHex = expiration.toString(16)
+    if (expirationHex.length % 2 === 1) {
+      expirationHex = '0' + expirationHex
+    }
+    expirationHex = expirationHex.match(/.{2}/g).reverse()
+    expirationHex.length = Math.min(expirationHex.length, 5)
+    expirationHex[expirationHex.length - 1] = '00'
+
+    return `76a97263a914${secretHash}8814${recipientAddress}67${expirationHex.length.toString(16).padStart(2, '0')}${expirationHex.join('')}b16d14${refundAddress}6888ac`
```
harshjv: I am not confident about adding such a code snippet in our library. Modification to this code snippet is not going to be a fun task. It is not expressive and is prone to error.

src/providers/ethereum/EthereumMetaMaskProvider.js
```diff
@@ -51,4 +51,24 @@ export default class EthereumMetaMaskProvider {
...
+    return `60${dataSizeEncoded}80600b6000396000f36020806000803760218160008060026048f17f\
```
harshjv, line 84: Same here. See my previous comment :)

src/index.js
harshjv: Yup! How about we use this library: https://www.npmjs.com/package/type-check

src/providers/bitcoin/BitcoinLedgerProvider.js
monokh: I see this as a temporary thing but what would be the alternative? I think we would either need to remove generating swaps from this library entirely or integrate bitcoin script / eth contract compilers into the library. I thought there were concerns around introducing huge libraries

harshjv: Can we do something like this?

```
const OP_MAP = {
  OP_DUP: '76',
  OP_HASH160: 'a9',
  OP_2SWAP: '72',
  OP_IF: '63',
  OP_EQUALVERIFY: '88',
  OP_ELSE: '67',
  OP_CHECKLOCKTIMEVERIFY: 'b1',
  OP_2DROP: '6d',
  OP_ENDIF: '68',
  OP_CHECKSIG: 'ac'
}

function compile (str) {
  const script = []
  const arr = str.split(' ')

  arr.forEach(v => {
    const code = OP_MAP[v]
    if (code) {
      script.push(code)
    } else {
      // check what is v and substitute proper value
    }
  })

  return script.join('')
}
```

monokh: We could do that, but are we then writing our own bitcoin script and ETH contract compilers? If that is not the end goal, it's hard to justify writing this if it's just for one specific script. I think if we decide to put compiling in right now, then we should do it properly. 

Once the script is set and tested there is no real room for modification if not adding an entirely new script. for readability opcodes can be closer to the hexes the represent in the form of comments

i3wgnit:
Should Swap really be a provider? I feel like it should be something else :confused: It doesn't make sense for us to put all the templates we might write as a provider right?

i3wgnit:
Should I be the one to introduce type check? Or should this be another PR?

src/providers/ethereum/EthereumSwapProvider.js
```diff
@@ -0,0 +1,109 @@
...
+      OP_AND: 0x16,
+      OP_JUMPI: 0x57,
+      OP_TIMESTAMP: 0x42,
+      OP_GT: 0x11,
+      OP_STOP: 0x00,
+      OP_JUMPDEST: 0x5b,
+      OP_SUICIDE: 0xff
+    }
+
+    return bytecode.match(/\S+/g)
```
monokh: I'm concerned about the precedent of introducing bytecode serialising. Let's say, next week we introduced a new script which had a new opcode that wasn't just as easy as `OP_BLA = 0x45`. Would we continue to write bytecode serialising in this library? I think if we are doing bytecode serialising in this project then we should be using a well tested library at that point. Therefore i prefer to not start with this problem until we really need to attend to it and my preference is for something very simple like: 
```
const changeP2PKHScript = [
      '76', // OP_DUP
      'a9', // OP_HASH160
      '14', // data size to be pushed
      changePubKeyHash, // <PUB_KEY_HASH>
      '88', // OP_EQUAL_VERIFY
      'ac' // OP_CHECKSIG
    ].join('')
```

monokh:
@i3wgnit i think it being inside the provider for now is fine. The providers can be really modular and that's a good thing. We only need swaps at this point. Once/if we have others, we can think about it.

monokh (changes requested):

monokh (approved):

monokh:
Feel free to merge @i3wgnit CI is failing but matt has fixed this

PR #16, opened 2018-07-26
Add crypto fns

matthewjablack:


package.json
```diff
@@ -91,7 +94,8 @@
     ],
     "ignore": [
       "dist/**",
-      "docs/**"
+      "docs/**",
+      "src/util/**"
```
harshjv: Why are we excluding this?

matthewjablack: Nice catch, was required when not including crypto libraries through npm. Removed.

src/index.js
```diff
@@ -228,6 +237,14 @@ export default class ChainAbstractionLayer {
     return addresses
   }
 
+  async getAddress () {
+    this._checkMethod('getAddresses')
```
harshjv: `getAddress` is checking for `getAddresses`?

src/index.js
```diff
@@ -247,6 +264,26 @@ export default class ChainAbstractionLayer {
...
+    const sha256Hashed = this.sha256(message)
+    const ripemd160Hashed = this.ripemd160(sha256Hashed)
+    return ripemd160Hashed
+  }
+
+  async sha256 (message) {
+    return sha256(message).toString()
+  }
+
+  async ripemd160 (message) {
```
harshjv: Can we add documentation here with new methods?

harshjv (changes requested):

i3wgnit (changes requested):
Could you create a property in cal for these crypto functions? So `CAL.crypto.hash160` along these lines.

matthewjablack:
Updated with your recommended changes @i3wgnit

src/index.js
matthewjablack: Yes because this function is using getAddresses and returning the first item in the array (primary address)

src/crypto.js
```diff
@@ -0,0 +1,53 @@
...
+   * @return {string} Returns the base58 of a string.
+   */
+  base58: basex(BASE58),
+
+  /**
+   * Get bech32 of message.
+   * @param {!string} message - any string.
+   * @return {string} Returns the bech32 of a string.
+   */
+  bech32: bech32,
```
i3wgnit: Why is this not `bech32,`?

src/index.js
```diff
@@ -24,6 +25,7 @@ export default class ChainAbstractionLayer {
     if (provider) {
       this.addProvider(provider)
     }
+    this.crypto = crypto
```
monokh: Why add it to the client as a property? If anyone wants it, it should just be exported on the module level and people can use/import as needed

src/crypto.js
```diff
@@ -0,0 +1,53 @@
...
+   * @return {string} Returns the bech32 of a string.
+   */
+  bech32,
+
+  /**
+   * Get hash160 of message.
+   * @param {!string} message - any string.
+   * @return {string} Returns the hash160 of a string.
+   */
+  hash160 (message) {
```
i3wgnit, line 28: crypto is not a class, this should be `hash160: (message) => {}`

src/crypto.js
```diff
@@ -0,0 +1,53 @@
...
+    const ripemd160Hashed = this.ripemd160(sha256Hashed)
+    return ripemd160Hashed
+  },
+
+  /**
+   * Get sha256 of message.
+   * @param {!string} message - any string.
+   * @return {string} Returns the sha256 of a string.
+   */
+  sha256 (message) {
```
i3wgnit, line 39: Idem

src/crypto.js
```diff
@@ -0,0 +1,53 @@
...
+    return sha256(message).toString()
+  },
+
+  /**
+   * Get ripemd160 of message.
+   * @param {!string} message - any string.
+   * @return {string} Returns the ripemd160 of a string.
+   */
+  ripemd160 (message) {
```
i3wgnit, line 48: idem

src/index.js
matthewjablack: Updated

i3wgnit (dismissed):
Its pretty good i like this :-)

src/crypto.js
matthewjablack: Updated

i3wgnit (approved):
Good stuff :-)

monokh:
Harsh's comments seem resolved so merging :)

PR #18, opened 2018-07-27
Embracing the new providers model

i3wgnit:


src/index.js
```diff
@@ -62,7 +62,7 @@ export default class ChainAbstractionLayer {
 
     const provider = findLast(this._providers, provider => isFunction(provider[method]), indexOfRequestor - 1)
 
-    if (!provider) {
+    if (provider == null) {
```
harshjv, line 65: It returns `undefined` if it finds nothing. Why are you comparing it loosely (`==`) with `null`?

src/Provider.js
```diff
@@ -2,4 +2,8 @@ export default class Provider {
   setClient (client) {
     this.client = client
   }
+
+  getMethod (method) {
+    return this.client.getProviderForMethod(method, this)[method]
```
harshjv: This won't work. Need to bind function to the provider first.

harshjv (changes requested):

src/Provider.js
i3wgnit: This works the first time around, however, adding a provider and using this method in it will remove the this object in the previously added provider.

This should go somewhere at least.

src/index.js
i3wgnit: This is how [StandardJs](https://standardjs.com/) says this should be done.
It says to `== null` to check for `null || undefined`.

i3wgnit: I quote
> Always use === instead of == – but obj == null is allowed to check null || undefined.

src/Provider.js
harshjv: Change it with this then.

```
getMethod (method) {
  const provider = this.client.getProviderForMethod(method, this)
  return provider[method].bind(provider)
}
```

examples/node/ethereum/multiple-providers.js
```diff
@@ -10,34 +11,34 @@ class X {
...
-    const res = provider.signMessage(message, from)
-    return 'Z' + ' ' + res
+    // const provider = this.client.getProviderForMethod('signMessage', this)
+    // const res = provider.signMessage(message, from)
+    return 'Z ' + this.getMethod('signMessage')(message, from)
   }
 }
 
 const ethereum = new ChainAbstractionLayer()
+ethereum.addProvider(new X())
```
matthewjablack: Change this to reflect chaining `addProvider`

examples/node/ethereum/multiple-providers.js
```diff
@@ -1,6 +1,7 @@
 const ChainAbstractionLayer = require('../../../')
+const Provider = ChainAbstractionLayer.Provider
 
-class X {
+class X extends Provider {
   setClient (client) {
```
matthewjablack: Remove all `setClient`

src/providers/bitcoin/BitcoinLedgerProvider.js
```diff
@@ -1,19 +1,15 @@
...
-import BitcoinRPCProvider from './BitcoinRPCProvider'
-
-export default class BitcoinLedgerProvider extends BitcoinRPCProvider {
-  constructor (uri, user, pass) {
-    super(uri, user, pass)
+export default class BitcoinLedgerProvider extends Provider {
+  constructor () {
+    super()
     this._ledgerBtc = false
     this._derivationPath = `44'/0'/0'/0`
```
matthewjablack, line 10: @harshjv are we sure we want this to always be the default. Does this use legacy bitcoin address

src/providers/ethereum/EthereumLedgerProvider.js
```diff
@@ -1,19 +1,15 @@
...
-import EthereumRPCProvider from './EthereumRPCProvider'
-
-export default class EthereumLedgerProvider extends EthereumRPCProvider {
-  constructor (uri, user, pass) {
-    super(uri, user, pass)
+export default class EthereumLedgerProvider extends Provider {
+  constructor () {
+    super()
     this._ledgerEth = false
     this._derivationPath = `44'/60'/0'/0'/0`
```
matthewjablack, line 10: @harshjv are we sure that this is the path we want for Ethereum everytime. Probably not as big of a deal since no segwit for Ethereum.

examples/node/ethereum/multiple-providers.js
```diff
@@ -1,43 +1,32 @@
 const ChainAbstractionLayer = require('../../../')
+const Provider = ChainAbstractionLayer.Provider
```
harshjv, line 2: Or you can just do; (for consistency)

    const { Provider } = ChainAbstractionLayer

examples/node/ethereum/multiple-providers.js
```diff
@@ -1,43 +1,32 @@
...
-    this.client = client
-  }
-
+class Z extends Provider {
   signMessage (message, from) {
-    const provider = this.client.getProviderForMethod('signMessage', this)
-    const res = provider.signMessage(message, from)
-    return 'Z' + ' ' + res
+    // const provider = this.client.getProviderForMethod('signMessage', this)
+    // const res = provider.signMessage(message, from)
```
harshjv: You can remove the comments.

harshjv (approved):

harshjv:
@i3wgnit LGTM (just need to remove comments)

examples/node/ethereum/multiple-providers.js
i3wgnit: CAL is used below.

PR #22, opened 2018-07-31
Bitcoin ledger send transaction

monokh:
very much a WIP at the moment. Trying to get it working first before thinking about how to structure it.
Still if any suggestions please add 🙂


src/providers/bitcoin/BitcoinLedgerProvider.js
```diff
@@ -21,6 +25,89 @@ export default class BitcoinLedgerProvider extends Provider {
     this._derivationPath = path
   }
 
+  async _getAddressDetails (address) {
+    return (await axios.get(`https://testnet.blockchain.info/balance?active=${address}&cors=true`)).data[address]
```
harshjv: `BitcoinLedgerProvider` is just a signing provider. It should't have network specific calls in it.

src/providers/bitcoin/BitcoinLedgerProvider.js
```diff
@@ -21,6 +25,89 @@ export default class BitcoinLedgerProvider extends Provider {
...
+      const addressDetails = await this._getAddressDetails(address.bitcoinAddress)
+      addressHasTransactions = addressDetails.n_tx > 0
+      if (addressHasTransactions) {
+        let utxos = await this._getUnspentTransactions(address.bitcoinAddress)
+        utxos = utxos.map((utxo) => ({
+          ...utxo,
+          path,
+          address: address.bitcoinAddress
+        }))
+        unspentInputs.push.apply(unspentInputs, utxos)
```
harshjv: I am confused why did you need `.apply`.

harshjv (changes requested):

src/providers/bitcoin/BitcoinLedgerProvider.js
monokh: `utxos` is an array. so this line is adding every element in `utxos` to `unspentInputs`

src/providers/bitcoin/BitcoinLedgerProvider.js
monokh: How would you structure this?
We cannot have other provider methods for this because these are bitcoin specific functions that are not able to be implemented in any other chain. 
A utility module? 
Base provider module? I cannot imagine that one helping the situation

src/providers/bitcoin/BitcoinLedgerProvider.js
harshjv: you can just do;

```
unspentInputs.push(...utxos)
```

🤷🏻‍♂️

harshjv (changes requested):

src/providers/bitcoin/BitcoinLedgerProvider.js
monokh: `[].push([])` is NOT the same as `[].push.apply([])`
Try this in node console to see the difference
```
let myarray = [1, 2, 3]
myarray.push([4, 5])
console.log(myarray)
```
compare it to 
```
let myarray = [1, 2, 3]
myarray.push.apply(myarray, [4, 5])
console.log(myarray)
```

harshjv: I got the point but we are already using ES6. Just use spread operator

```
unspentInputs.push(...utxos)
```

monokh: Good idea 👍 will do
waiting for further comments to implement all in one go

src/providers/bitcoin/BitcoinLedgerProvider.js
harshjv: In this case, this will be `BitcoinLedgerTestnetBlockchainProvider`!

You can move everything except signing methods to `BitcoinRPCProvider` or create a new provider `BitcoinBlockchainProvider`.

You will be still able to use all the methods.

harshjv (changes requested):

src/providers/bitcoin/BitcoinLedgerProvider.js
monokh: I'm not quite following,
None of the methods here are to do with RPC, if we had a BitcoinBlockchainProvider, which methods would you carry over and how would you communicate between the providers? Shouldn't the communication of providers be only on normalized methods? there aren't any here that are not bitcoin specific

monokh:
@harshjv i actually realised that even putting into another module will have a lot of thinking to do on how to structure it. So I went with the minimal changes so we can unblock everyone for now.

PR #23, opened 2018-07-31
JsonRpcProvider base class

harshjv:
Currently, `jsonrpc` specific stuff is implemented in both `BitcoinRPCProvider` and `EthereumRPCProvider`. Now that we are supporting multiple-providers, moving `jsonrpc` stuff to `JsonRpcProvider` and `BitcoinRPCProvider/EthereumRPCProvider` leverages `JsonRpcProvider` for `jsonrpc` calls and checks.

`JsonRpcProvider` extends `Provider` base class and exposes `rpc` method.

So this is how provider stack will look like:

    const ChainAbstractionLayer = require('chainabstractionlayer')
    const { EthereumLedgerProvider, EthereumRPCProvider } = ChainAbstractionLayer.providers.ethereum

    const ethereum = new ChainAbstractionLayer()
    ethereum.addProvider(new EthereumLedgerProvider()) // no-dependency provider[0]
    ethereum.addProvider(new EthereumRPCProvider('http://localhost:8545')) // provider[1] can use methods added by provider[0]
    ethereum.addProvider(new EthereumLedgerProvider()) // provider[2] can use methods added by provider[1]


src/providers/JsonRpcProvider.js
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    if (!has(data, 'result')) {
+      throw new RpcError(-32700, 'Missing `result` on the RPC call result')
+    }
+
+    return data.result
+  }
+
+  rpc (method, ...params) {
```
i3wgnit, line 51: Just me being curious here, whats the difference between `_someMethod` and `someMethod`. Why the underscore sometimes?

i3wgnit (dismissed):
I like the direction this is going :+1:

src/providers/RpcError.js
```diff
@@ -0,0 +1,36 @@
+import { STATUS_CODES } from 'http'
+import StandardError from './StandardError'
+
+export default class RpcError extends StandardError {
+  constructor (code, msg, props = {}) {
+    if (typeof code !== 'number') {
+      throw new TypeError(`Non-numeric HTTP code`)
+    }
+
+    if (typeof msg === 'object' && msg !== null) {
```
i3wgnit, line 10: Shouldn't this be `msg != null` ?

package.json
```diff
@@ -38,7 +38,8 @@
     "crypto-js": "^3.1.9-1",
     "json-bigint": "^0.3.0",
     "lodash": "^4.0.0",
-    "regenerator-runtime": "^0.12.0"
+    "regenerator-runtime": "^0.12.0",
+    "standard-error": "^1.1.0"
```
i3wgnit, line 42: How big is this?

i3wgnit (changes requested):
I like the direction :+1:

package.json
harshjv: screen shot 2018-08-02 at 11 43 53 am

src/providers/JsonRpcProvider.js
harshjv: Methods starting with `_` are private methods. (since in js there is no way of making a method private, using `_` as prefix, document generators mark them as `private` in docs.)

Since `rpc` method is explicitly exposed by the `JsonRpcProvider`, it has no `_`.

src/providers/RpcError.js
harshjv: No. `null` is an `object`. checking with `!=` will also check for `undefined` which we don't want since first check is for `object`.

Umm, Instead, we can only do `msg != null`. Is that what you meant?

src/providers/JsonRpcProvider.js
```diff
@@ -0,0 +1,57 @@
...
+    return this._axios.post(
+      '/',
+      this._prepareRequest(method, params)
```
monokh, line 54: does this need it's own method? Would be easier to read with the content of the method just here since `_prepareRequest` isn't complicated. Unless it can be more complicated in the future

src/providers/JsonRpcProvider.js
```diff
@@ -0,0 +1,57 @@
...
+  constructor (uri, username, password) {
+    super()
+
+    this._axios = axios.create({
+      baseURL: uri,
+      responseType: 'text',
+      transformResponse: undefined // https://github.com/axios/axios/issues/907
+    })
+
+    if (username || password) {
```
monokh, line 20: should both username and password be required for this? Is there a case where an rpc node will have a username but no password or the opposite?

monokh (commented):
Some small comments but no objection to merging in this state

src/providers/RpcError.js
i3wgnit: Oh, I didn't know that null was type object.

i3wgnit: Why not make it extend props instead of replace then?

i3wgnit (approved):

src/providers/JsonRpcProvider.js
harshjv: Nope.

That's how http basic auth works. https://github.com/jshttp/basic-auth#with-vanilla-nodejs-http-server

src/providers/JsonRpcProvider.js
harshjv: It is this way for many reasons. First is, if you want to add prefix to request `id` or do some other modification, you just need to overwrite `_prepareRequest`.

src/providers/JsonRpcProvider.js
monokh: not sure if I explained properly, i know basic auth but does it make sense to provide only a password because in this case basic auth will be added. Anyway, this isn't anything very important

monokh (approved):

monokh:
Happy to merge when conflicts resolved

PR #30, opened 2018-08-06
Errors, documentation for errors and updated examples

### LINKED ISSUE #24: Subclass errors
Currently, `Error` is thrown when something goes wrong. We should start using subclasses of `Error` like `TypeError` etc.

harshjv:
Fixes #24 

- [x] Subclass errors
- [x] Add documentation for errors and methods
- [x] Update examples
- [x] Update readme
- [x] Update tests

### DIFF
--- a/examples/node/bitcoin/ledger.js
+++ b/examples/node/bitcoin/ledger.js
@@ -3,10 +3,9 @@ const { BitcoinLedgerProvider } = ChainAbstractionLayer.providers.bitcoin
 
 const prettyPrintJson = (val) => console.log(JSON.stringify(val, null, 2))
 
-const bitcoinLedgerProvider = new BitcoinLedgerProvider('http://localhost:8080', 'bitcoin', 'local321')
-const bitcoin = new ChainAbstractionLayer(bitcoinLedgerProvider)
+const bitcoin = new ChainAbstractionLayer()
+bitcoin.addProvider(new BitcoinLedgerProvider())
 
 ;(async () => {
-  prettyPrintJson(await bitcoin.generateBlock(1))
   prettyPrintJson(await bitcoin.signMessage('hello world'))
 })()
--- a/src/errors.js
+++ b/src/errors.js
@@ -1,0 +1,15 @@
+import BaseError from 'standard-error'
+
+function createError (name) {
+  class NewError extends BaseError {}
+  NewError.prototype.name = name
+  return NewError
+}
+
+export const StandardError = createError('StandardError')
+export const InvalidProviderError = createError('InvalidProviderError')
+export const DuplicateProviderError = createError('DuplicateProviderError')
+export const NoProviderError = createError('NoProviderError')
+export const UnsupportedMethodError = createError('UnsupportedMethodError')
+export const UnimplementedMethodError = createError('UnimplementedMethodError')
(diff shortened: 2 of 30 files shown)


src/errors.js
```diff
@@ -0,0 +1,15 @@
+import BaseError from 'standard-error'
```
monokh, line 1: I didn't notice `standard-error` being added to `package.json` i think i found it part of `babel/core`. Is it good practice to rely on transitive dependencies or better to add it as a direct dependency of the project? Just wondering

src/index.js
```diff
@@ -5,42 +5,47 @@ import * as Ajv from 'ajv'
...
 
 export default class ChainAbstractionLayer {
   /**
    * ChainAbstractionLayer client
-   * @param {object} [provider] - The provider instance.
-   * @param {string} [version] - Version string
+   * @param {string} [version] - Minimum blockchain node version to support
    */
-  constructor (provider, version) {
+  constructor (version) {
```
monokh: Is the version used? do we need to keep it?

monokh (changes requested):
LGTM, just some small questions

src/errors.js
harshjv: `standard-error` was already in our `package.json`. I was earlier using it for `RpcError`.

src/index.js
harshjv: Adding a provider directly in constructor has very limited use cases because one single provider cannot do all the tasks. Should we keep the old version?

src/errors.js
monokh: sure is 👍  cool

src/index.js
monokh: I see your point about the provider and i agree, but I'm asking about the `version` 😄

monokh: I guess nothing to do with this PR actually so i'm happy to address that later

monokh (approved):

src/index.js
harshjv: Oh. Yes, `version` is used. See https://github.com/ConsenSys/chainabstractionlayer/blob/master/src/index.js#L70-L74

examples/node/bitcoin/ledger.js
```diff
@@ -3,10 +3,9 @@ const { BitcoinLedgerProvider } = ChainAbstractionLayer.providers.bitcoin
 
 const prettyPrintJson = (val) => console.log(JSON.stringify(val, null, 2))
 
-const bitcoinLedgerProvider = new BitcoinLedgerProvider('http://localhost:8080', 'bitcoin', 'local321')
-const bitcoin = new ChainAbstractionLayer(bitcoinLedgerProvider)
+const bitcoin = new ChainAbstractionLayer()
+bitcoin.addProvider(new BitcoinLedgerProvider())
```
matthewjablack, line 7: Is there a reason for removing ability to add providers through constructor. I see that there is more consistency in adding providers rather than initializing and then adding providers (since most users will be adding multiple providers), but should we still provide option to add provider through constructor?

harshjv: We can do that too. I removed it because, in all of our examples, we are using multiple providers. For now, let's keep it. We will discuss this again if something comes up.

matthewjablack (approved):
Thanks for that small change @harshjv 

LGTM 🎉

PR #33, opened 2018-08-07
Export different modules

### LINKED ISSUE #25: Export different modules
Currently, `ChainAbstractionLayer` is exported by default and `Provider`, `providers`, `schemas`, `crypto` are added statically to `ChainAbstractionLayer`. This makes it hard to only import specific modules of our library.

## index.js

```
import ChainAbstractionLayer from './ChainAbstractionLayer'
import Provider from './Provider'
import crypto from './crypto'
import providers from './providers'
import BlockSchema from './schema/Block.json'
import TransactionSchema from './schema/Transaction.json'

export {
  ChainAbstractionLayer,
  Provider,
  crypto,
  providers,
  schemas: {
    Block: BlockSchema,
    Transaction: TransactionSchema
  }
}
```

harshjv:
The aim of this PR is to export separate modules instead of exposing them statically on `ChainAbstractionLayer` object.

Essentially;

```diff
-const ChainAbstractionLayer = require('chainabstractionlayer')
-const { crypto } = ChainAbstractionLayer
-const client = new ChainAbstractionLayer()
+const { Client, crypto } = require('chainabstractionlayer')
+const client = new Client()
```

Fixes #25


monokh (approved):

webpack.browser.config.js
```diff
@@ -34,11 +34,7 @@ module.exports = {
     path: path.resolve(__dirname, 'dist'),
     filename: path.basename(pkg.browser),
     library: 'ChainAbstractionLayer',
-    libraryTarget: 'umd',
-    libraryExport: 'default'
-  },
-  node: {
-    fs: 'empty'
```
matthewjablack, line 41: I'm guessing `fs: 'empty'` is no longer needed since we don't need `bcrypto` anymore.

matthewjablack (approved):
Just one small question. Other than that LGTM 🎉

webpack.browser.config.js
harshjv: Yep!

PR #38, opened 2018-08-08
Convert getAddresses to return object and add coinType

matthewjablack:
This PR adds global network config i.e. 

```
export default {
  bitcoin: {
    pubKeyHash: '00',
    scriptHash: '05',
    coinType: '0',
    explorerUrl: 'https://blockchain.info'
  },
  testnet: {
    pubKeyHash: '6F',
    scriptHash: 'C4',
    coinType: '1',
    explorerUrl: 'https://testnet.blockchain.info'
  },
  litecoin: {
    pubKeyHash: '30',
    scriptHash: '32',
    coinType: '2'
  },
  ethereum: {
    coinType: '60'
  },
  ethereum_classic: {
    coinType: '61'
  }
}

```

as well as segwit support for `getAddresses` with `coinType` specification (https://github.com/satoshilabs/slips/blob/master/slip-0044.md)

### DIFF
--- a/examples/browser/bitcoin/ledger.html
+++ b/examples/browser/bitcoin/ledger.html
@@ -11,9 +11,9 @@
   <p><code>For errors and logs, check console</code></p>
   <script>
   /* global $, ChainAbstractionLayer */
-  const { Client, providers } = ChainAbstractionLayer
+  const { Client, providers, networks } = ChainAbstractionLayer
   const bitcoin = new Client()
-  bitcoin.addProvider(new providers.bitcoin.BitcoinLedgerProvider())
+  bitcoin.addProvider(new providers.bitcoin.BitcoinLedgerProvider({ network: networks.bitcoin }))
   bitcoin.getAddresses().then(addresses => {
     const from = addresses[0]
     $('#address').text(from)
--- a/src/Client.js
+++ b/src/Client.js
@@ -290,10 +290,10 @@ export default class Client {
    *  of accounts.
    *  Rejects with InvalidProviderResponseError if provider's response is invalid.
    */
-  async getAddresses () {
+  async getAddresses (segwit = false) {
     const provider = this.getProviderForMethod('getAddresses')
 
-    const addresses = await provider.getAddresses()
+    const addresses = await provider.getAddresses(segwit)
 
     if (!isArray(addresses)) {
       throw new InvalidProviderResponseError('Provider returned an invalid response')
@@ -366,8 +366,8 @@ export default class Client {
    */
   async generateSecret (message) {
     const addresses = await this.getAddresses()
-    const from = addresses[0]
-    const signedMessage = await this.signMessage(message, from)
+    const path = addresses[0].path
+    const signedMessage = await this.signMessage(message, path)
     const secret = hash160(signedMessage)
     return secret
   }
--- a/src/Client.js
+++ b/src/Client.js
@@ -364,9 +364,7 @@ export default class Client {
    * @param {!string} message - Message to be used for generating secret.
    * @return {Promise<string, null>} Resolves with a secret.
    */
-  async generateSecret (message) {
-    const addresses = await this.getAddresses()
-    const path = addresses[0].path
+  async generateSecret (message, path) {
     const signedMessage = await this.signMessage(message, path)
     const secret = hash160(signedMessage)
     return secret
--- a/src/Client.js
+++ b/src/Client.js
@@ -290,10 +290,10 @@ export default class Client {
    *  of accounts.
    *  Rejects with InvalidProviderResponseError if provider's response is invalid.
    */
-  async getAddresses (segwit = false) {
+  async getAddresses (config) {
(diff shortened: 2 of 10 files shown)


examples/browser/bitcoin/ledger.html
```diff
@@ -15,7 +15,7 @@ <h1>CAL + Ledger: Sign a message (Bitcoin)</h1>
   const bitcoin = new Client()
   bitcoin.addProvider(new providers.bitcoin.BitcoinLedgerProvider())
   bitcoin.getAddresses().then(addresses => {
-    const from = addresses[0]
+    const from = addresses[0].bitcoinAddress
```
harshjv: Why is this changed? it should return an array of addresses. Not an object.

src/providers/ethereum/EthereumLedgerProvider.js
```diff
@@ -3,11 +3,14 @@ import Provider from '../../Provider'
 import Transport from '@alias/ledger-transport'
 import LedgerEth from '@ledgerhq/hw-app-eth'
 
+import { networks } from './EthereumUtil'
+
 export default class EthereumLedgerProvider extends Provider {
-  constructor () {
+  constructor (chain = 'ethereum') {
```
harshjv: Um, why?

Edit: saw EthereumUtil.js. Can we move network object to the root? Possibly, we can make symbol/const for the networks. i.e.

### network.js
```js
export default {
  BITCOIN: {
    TESTNET: {},
    MAINNET: {},
  },
  ETHEREUM: {
    RINKEBY: {},
    MAINNET: {},
  }, ...
}
```

and then,

```js
import { network } from 'chainabstractionlayer'
...
client.addProvider(new EthereumLedgerProvider({ network: network.ETHEREUM.MAINNET }))
```

We can enforce this signature for each provider, that it expects network in either an object or a value. What do you think? This way, we will be housing all network specific config in a single file.

harshjv (changes requested):

examples/browser/bitcoin/ledger.html
matthewjablack: The majority of the time when you are getting addresses you are also looking to get the balance of that address as well. Made sense to include balance information, as well as utxo info for bitcoin chain.

src/providers/ethereum/EthereumLedgerProvider.js
matthewjablack: @harshjv I agree with your suggestion to move everything to one file. However I think each chain (including testnets) should be their own `chain` in this config. 

Looking at https://github.com/satoshilabs/slips/blob/master/slip-0044.md we can see that `Bitcoin Testnet` is a separate `coinType` than `Bitcoin` (similar with `Ethereum` and `Ethereum Classic`). However, `Rinkeby`, `Koven`, `Mainnet` all have the same `coinType`. So we can either specify `chainType` i.e. `Bitcoin` or `Ethereum` and then the underlying chain, or have that at root level. Thoughts?

examples/browser/bitcoin/ledger.html
matthewjablack: Updated to return array of addresses.

matthewjablack:
@harshjv PR updated with your suggestions

harshjv (approved):
LGTM

i3wgnit (approved):
:+1: long time overdue :) we now have access to all bitcoin forks

src/Client.js
```diff
@@ -290,10 +290,10 @@ export default class Client {
    *  of accounts.
    *  Rejects with InvalidProviderResponseError if provider's response is invalid.
    */
-  async getAddresses () {
+  async getAddresses (segwit = false) {
```
harshjv: `segwit` is a bitcoin specific term. is it okay to leave it in client.js?

harshjv: can we pass a `config` object instead? i.e. `{ segwit }`

```
async getAddresses (config = {})
```

and when fetching segwit addresses from bitcoin client, we can do;

```
client.getAddresses({ segwit: true })
```

Any thoughts?

src/networks.js
```diff
@@ -0,0 +1,25 @@
+export default {
+  bitcoin: {
+    pubKeyHash: '00',
+    scriptHash: '05',
+    coinType: '0',
+    explorerUrl: 'https://blockchain.info'
+  },
+  testnet: {
```
harshjv: Suppose this is accessed as `networks.testnet` somewhere in the codebase. How can a dev/user/auditor infer that this `networks.testnet` is Bitcoin testnet and not Ethereum testnet or any other coin's testnet? Because there's also a `networks.bitcoin`.

Any thoughts?

harshjv (changes requested):

src/networks.js
matthewjablack: Perhaps we could specify the parent chain within the config

```
testnet: {
  parentChain: 'bitcoin'
}
```

Thoughts?

harshjv (approved):

PR #45, opened 2018-08-12
Convert value to hex

harshjv:
- [x] Fixes `sendTransaction` in EthereumMetaMaskProvider

### DIFF
--- a/src/providers/ethereum/EthereumMetaMaskProvider.js
+++ b/src/providers/ethereum/EthereumMetaMaskProvider.js
@@ -2,6 +2,7 @@ import Provider from '../../Provider'
 
 import { isFunction } from 'lodash'
 import { formatEthResponse, ensureEthFormat } from './EthereumUtil'
+import { BigNumber } from 'bignumber.js'
 
 export default class EthereumMetaMaskProvider extends Provider {
   constructor (metamaskProvider) {
@@ -51,7 +52,7 @@ export default class EthereumMetaMaskProvider extends Provider {
   }
 
   async sendTransaction (from, to, value, data) {
-    value = Number(value).toString(16)
+    value = BigNumber(value).toString(16)
 
     const tx = {
       from, to, value, data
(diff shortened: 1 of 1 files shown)


src/providers/ethereum/EthereumMetaMaskProvider.js
```diff
@@ -51,6 +51,8 @@ export default class EthereumMetaMaskProvider extends Provider {
   }
 
   async sendTransaction (from, to, value, data) {
+    value = Number(value).toString(16)
```
monokh: I think as a standard, we should be using `BigNumber` anywhere we represent values, there may be some unexpected results when numbers go over the [MAX_SAFE_INT](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER)

monokh (changes requested):

src/providers/ethereum/EthereumMetaMaskProvider.js
harshjv: This hotfix is only for `sendTransaction` which is blocking https://github.com/ConsenSys/liquality-atomic-swap/pull/7.

For numbers, I do agree to go with a library like BigInt (cool because native support is coming) or BigNumber.js. We will explore more edge cases when we dive into testing :)

monokh: Going over MAX_SAFE_INT itself could mean the value would be broken so since we are here i thought we might as well make the right decisions. BigNumber if i remember right is already included in CAL, so only need to replace `Number` with `BigNumber`

monokh: also i don't think this blocks https://github.com/ConsenSys/liquality-atomic-swap/pull/7 per se.  that PR can just assume that this will be fixed

monokh (approved):
LGTM

PR #46, opened 2018-08-16
Add refund and redeem to swap

i3wgnit:
### Description

Implements `refundSwap` and `redeemSwap` methods in BitcoinSwapProvider and EthereumSwapProvider.

### Parts

- [x] Implement `redeemSwap`
- [x] Implement `refundSwap`
- [x] Implement Relevant tests

### Considerations

- Rename for `redeemSwap` to `generateRedeemSwap`
- Rename `refundSwap` to `generateRefundSwap`
- Rename `generateSwap` to `generateInitSwap`


src/providers/bitcoin/BitcoinSwapProvider.js
```diff
@@ -2,36 +2,72 @@ import Provider from '../../Provider'
 import { addressToPubKeyHash } from './BitcoinUtil'
 
 export default class BitcoinSwapProvider extends Provider {
+  _padHexStart (hex, length) {
```
monokh: can this be in a util somewhere instead?

monokh:
@i3wgnit now that we are doing quite a bit of script and bytecode serialisation, i'm wondering, are there any small libraries that are well tested that can help us do this? Feel like this side of it is getting quite hard to read. Not a necessity but just to get us thinking

i3wgnit:
Bitcoinjslib imports a few small libraries here and there. I am not too familiar with them, but I have gone to them and tried to reverse engineer a few processes.

src/providers/bitcoin/BitcoinSwapProvider.js
i3wgnit: We can put this in crypto?

harshjv: IMO leave it here for now. we will rearrange it when we reuse it somewhere else.

harshjv (commented):
Waiting for the tests

PR #48, opened 2018-08-23
Fix getAddresses

i3wgnit:
### Description
Fix getAddress and implement getUsedAddresses and getUnusedAddresses

### Parts

- [x] Refactor `_getSpendingDetails`
- [x] Implement a working `getAddresses`
- [x] Implement `getUsedAddresses`
- [x] Implement `getUnusedAddresses`
- [x] Make an interface for the previous two bullet points in the client

PR #49, opened 2018-08-23
Implement getBalance for Bitcoin

i3wgnit:
### Description

Implement getBalance for BitcoinLedgerProvider
First merge #48 before this one.

### Parts

- [x] Implement `getBalance` method
- [x] Make an interface for the previous bullet point in the `Client`

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -288,7 +288,7 @@ export default class Client {
    * Get the balance of an account given its addresses.
    * @param {string[]} [addresses] - A list of addresses.
    * @return {Promise<number, InvalidProviderResponseError>} If addresses is given,
-   *  returns the cummulative balance of the given addresses. Otherwise return the balance
+   *  returns the cumulative balance of the given addresses. Otherwise returns the balance
    *  of the account that the signing provider controls.
    *  Rejects with InvalidProviderResponseError if provider's response is invalid.
    */
--- a/src/Client.js
+++ b/src/Client.js
@@ -289,7 +289,7 @@ export default class Client {
    * @param {string[]} [addresses] - A list of addresses.
    * @return {Promise<number, InvalidProviderResponseError>} If addresses is given,
    *  returns the cumulative balance of the given addresses. Otherwise returns the balance
-   *  of the account that the signing provider controls.
+   *  of the addresses that the signing provider controls.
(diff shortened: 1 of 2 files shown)


src/providers/bitcoin/BitcoinLedgerProvider.js
```diff
@@ -121,6 +121,13 @@ export default class BitcoinLedgerProvider extends Provider {
     return buffer.reverse() // Amount needs to be little endian
   }
 
+  async getBalance (config = { segwit: false }) {
```
monokh: i find it quite un normalized that the functions accept a config like that... the same interface won't exist for ethereum. The correct way i believe should be to set this flag on the provider constructor, but that's a different issue that doesn't need to be addressed here.

src/providers/bitcoin/BitcoinLedgerProvider.js
```diff
@@ -121,6 +121,13 @@ export default class BitcoinLedgerProvider extends Provider {
...
+    await this._connectToLedger()
+
+    const { usedAddresses } = await this._getSpendingDetails(config)
+    return usedAddresses.reduce((acc, detail) => acc + detail.value, 0)
```
monokh: This type of thing is a good candidate for a unit test. Mock _connectToLedger and _getSpendingDetails and see if the function does what you expect. Again different issue 🙂 feel free to address it whenever in your priorities

monokh (approved):
LGTM, minor comments but nothing that can't be done separately

monokh (changes requested):
looks fine but again can you include an example so we can see how to use this? similarly for the ethereum version

src/Client.js
```diff
@@ -284,6 +284,40 @@ export default class Client {
...
+    if (addresses != null) {
+      if (!isArray(addresses)) {
+        throw new TypeError('Addresses should be an array')
+      }
+
+      if (!addresses.every(isString)) {
+        throw new TypeError('All addresses should be strings')
+      }
+
+      if (!addresses.every(addr => (/^[A-Fa-f0-9]+$/.test(addr)))) {
```
matthewjablack, line 307: Doesn't this need to be removed since bitcoin addresses are not hex strings?

PR #50, opened 2018-08-29
Fix sendtransaction btcledger

i3wgnit:
### Description

Quick fix for sendTransaction in BitcoinLedgerProvider.
First merge #48 and #49 

### Parts

- [x] Generalized `to` field in btc, allows for both p2pkh and p2sh through a helper function called `_generateScript`
- [x] Modified order of arguments for sendTransaction
- [x] Removed the fake promise of sending to multiple parties in bitcoin
- [x] Added examples in examples/browser/bitcoin/ledger-send-transaction.html

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -388,19 +388,19 @@ export default class Client {
 
   /**
    * Send a transaction to the chain
-   * @param {!string} from - The address identifier for the sender.
-   * @param {!string} to - The address identifier for the receiver.
+   * @param {string} to - The address identifier for the receiver.
    * @param {!number} value - Number representing the amount associated with.
    *  the transaction.
    * @param {!string} [data] - Optional data to send with the transaction.
+   * @param {!string} [from] - Optional address identifier for the sender.
    * @return {Promise<string, InvalidProviderResponseError>} Resolves with an identifier for
    *  the broadcasted transaction.
    *  Rejects with InvalidProviderResponseError if provider's response is invalid.
    */
-  async sendTransaction (from, to, value, data) {
+  async sendTransaction (to, value, data, from) {
     const provider = this.getProviderForMethod('sendTransaction')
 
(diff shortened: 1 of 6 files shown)


src/providers/bitcoin/BitcoinLedgerProvider.js
```diff
@@ -177,23 +182,56 @@ export default class BitcoinLedgerProvider extends Provider {
...
+      ].join('')
+    } else {
+      throw new Error('Not a valid address:', address)
+    }
+  }
+
+  async sendTransaction (to, value, data, from) {
     await this._connectToLedger()
+    let receivingAddress
+    if (to == null) {
```
monokh: if `to` can also be empty, i feel like the order from, to, value, data is more natural.

monokh (changes requested):
This looks fine to me but can you update the examples? The current send transaction examples will be broken and they should demonstrate how to send a p2sh transaction. adding this would mean that we have a place to refer to on how to do this type of transaction as well as a place to test that it works

src/Client.js
```diff
@@ -303,10 +303,6 @@ export default class Client {
       if (!addresses.every(isString)) {
         throw new TypeError('All addresses should be strings')
       }
-
-      if (!addresses.every(addr => (/^[A-Fa-f0-9]+$/.test(addr)))) {
```
matthewjablack, line 307: Why are we removing this. Is this not the case for ethereum anymore?

matthewjablack (approved):
Just one small comment, other than that LGTM 🎉

src/Client.js
i3wgnit: Because addresses are not hex in bitcoin :sad:

matthewjablack: Ah of course 😢

monokh (approved):
LGTM!

PR #52, opened 2018-08-30
GetBalance on Ethereum

i3wgnit:
### Description

Implement `getBalance` following #49 's documentation.
That is, the function takes in a list of addresses and gives a cumulative balance of all addresses.

### Parts

- [x] Add getBalance to RPC provider
- [ ] Add getBalance to MetaMask provider


matthewjablack (approved):
@monokh were talking, and I don't think we need getBalance in MetaMaskProvider. The reason for that is MetaMaskProvider should be focused on signing, and getting addresses, rather than any broadcasting or getting data.

monokh (approved):
Yes as discussed all we will need is the rpc implementations of get balance.

PR #56, opened 2018-09-03
Swap creation fix

monokh:
### Description

Small fixes to make the initiation of swap work

### Parts

- [x] Removed hex validation for generate swap - only ethereum addresses are in hex
- [x] Fixed example

### DIFF
--- a/examples/browser/bitcoin/ledger-send-transaction.html
+++ b/examples/browser/bitcoin/ledger-send-transaction.html
@@ -57,7 +57,7 @@
       $('#sendScript').click(() => {
         // pay to script:
         // null, value, script
-        bitcoin.sendTransaction(null, $('#scriptValue').val(), parseInt($('#toScript').val())).then(result => {
+        bitcoin.sendTransaction(null, parseInt($('#scriptValue').val()), $('#toScript').val()).then(result => {
           $('#result').text(JSON.stringify(result, null, 2))
         })
       })
--- a/src/Client.js
+++ b/src/Client.js
@@ -460,14 +460,6 @@ export default class Client {
       throw new TypeError('Secret hash should be a string')
     }
 
-    if (!(/^[A-Fa-f0-9]+$/.test(recipientAddress))) {
-      throw new TypeError('Recipient address should be a valid hex string')
-    }
-
-    if (!(/^[A-Fa-f0-9]+$/.test(refundAddress))) {
-      throw new TypeError('Refund address should be a valid hex string')
-    }
-
     if (!(/^[A-Fa-f0-9]+$/.test(secretHash))) {
       throw new TypeError('Secret hash should be a valid hex string')
     }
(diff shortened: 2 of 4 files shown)


examples/browser/bitcoin/ledger-send-transaction.html
```diff
@@ -57,7 +57,7 @@ <h3>Pay to custom script</h3>
       $('#sendScript').click(() => {
         // pay to script:
         // null, value, script
-        bitcoin.sendTransaction(null, $('#scriptValue').val(), parseInt($('#toScript').val())).then(result => {
+        bitcoin.sendTransaction(null, parseInt($('#scriptValue').val()), $('#toScript').val()).then(result => {
```
harshjv, line 60: this will fail for floats. e.g. 0.1 ETH

src/Client.js
```diff
@@ -460,14 +460,6 @@ export default class Client {
       throw new TypeError('Secret hash should be a string')
     }
 
-    if (!(/^[A-Fa-f0-9]+$/.test(recipientAddress))) {
```
harshjv, line 463: nice catch!

src/providers/bitcoin/BitcoinLedgerProvider.js
```diff
@@ -216,10 +216,10 @@ export default class BitcoinLedgerProvider extends Provider {
     await this._connectToLedger()
 
     let receivingAddress
-    if (to == null) {
+    if (data) {
```
harshjv, line 219: Nice catch, again!

src/providers/bitcoin/BitcoinUtil.js
```diff
@@ -30,7 +30,7 @@ function compressPubKey (pubKey) {
 function pubKeyToAddress (pubKey, network, type) {
   pubKey = ensureBuffer(pubKey)
   const pubKeyHash = hash160(pubKey)
-  const addr = this.pubKeyHashToAddress(pubKeyHash, network, type)
+  const addr = pubKeyHashToAddress(pubKeyHash, network, type)
```
harshjv, line 33: Another one! Thanks Mo!

examples/browser/bitcoin/ledger-send-transaction.html
harshjv: you can change it to `parseFloat` or use `Number`

harshjv (changes requested):
Just a small modification required! Other than that, LGTM

examples/browser/bitcoin/ledger-send-transaction.html
monokh: I believe the interface for `sendTransaction` accepts the smallest denomination of the amount. so 1 here would be 1 satoshi or wei. So I think we shouldn't be parsing as floats for the example unless we were doing some conversion from the various denominations

monokh:
comments have been addressed and don't see anything really contentious here so merging

PR #57, opened 2018-09-05
Add getBalance to MetaMask Provider

matthewjablack:
### Description

Add getBalance to MetaMaskProvider

- [ ] getBalance MetaMaskProvider

### DIFF
--- a/src/providers/ethereum/EthereumMetaMaskProvider.js
+++ b/src/providers/ethereum/EthereumMetaMaskProvider.js
@@ -41,6 +41,13 @@ export default class EthereumMetaMaskProvider extends Provider {
     })
   }
 
+  async getBalance () {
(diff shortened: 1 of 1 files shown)


src/providers/ethereum/EthereumMetaMaskProvider.js
```diff
@@ -41,6 +41,13 @@ export default class EthereumMetaMaskProvider extends Provider {
     })
   }
 
+  async getBalance () {
```
harshjv, line 44: Addresses should be passed explicitly so we can also query balance for other addresses.

harshjv (changes requested):

monokh:
As discussed with @mattBlackDesign rpc provider is sufficient to implement getbalance (already done). hence closing this

PR #58, opened 2018-09-06
Swap example

monokh:
### Description

Creates an e2e example for btc<->eth swap, as well as the required fixes to normalise the interface.

Initially only generation and funding

- [x] Add an example for swap functionality - generating and funding a swap
- [x] Fix metamask send transaction to make the right call as well as normalise the inputs and outputs

### DIFF
--- a/examples/browser/bitcoin/ledger-send-transaction.html
+++ b/examples/browser/bitcoin/ledger-send-transaction.html
@@ -29,7 +29,7 @@
     /* global $, ChainAbstractionLayer */
     const { Client, providers, networks } = ChainAbstractionLayer
     const bitcoin = new Client()
-    bitcoin.addProvider(new providers.bitcoin.BitcoinRPCProvider('http://localhost:8080', 'bitcoin', 'local321'))
+    bitcoin.addProvider(new providers.bitcoin.BitcoinRPCProvider('https://localhost:18332', 'bitcoin', 'local321'))
     bitcoin.addProvider(new providers.bitcoin.BitcoinLedgerProvider({ network: networks.bitcoin_testnet }))
 
     let addresses, balance, newAddresses
@@ -44,7 +44,7 @@
 
       bitcoin.getUnusedAddresses().then(unusedAddrs => {
         newAddresses = unusedAddrs
-        $('#newAddress').text(newAddresses[1])
+        $('#newAddress').text(newAddresses[0])
       })
 
       $('#sendAddr').click(() => {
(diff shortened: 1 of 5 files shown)


examples/browser/bitcoin/ledger-send-transaction.html
```diff
@@ -29,7 +29,7 @@ <h3>Pay to custom script</h3>
     /* global $, ChainAbstractionLayer */
     const { Client, providers, networks } = ChainAbstractionLayer
     const bitcoin = new Client()
-    bitcoin.addProvider(new providers.bitcoin.BitcoinRPCProvider('http://localhost:8080', 'bitcoin', 'local321'))
+    bitcoin.addProvider(new providers.bitcoin.BitcoinRPCProvider('https://localhost:18332', 'bitcoin', 'local321'))
```
matthewjablack, line 32: Wasn't part of the reason we kept this as a different port than 18332 was because there was the requirement to use `easy-cors-proxy` (https://github.com/i3wgnit/easy-cors-proxy) to avoid cors issues.

examples/swap/swap-e2e.html
```diff
@@ -0,0 +1,114 @@
+<html>
```
matthewjablack, line 1: awesome!!!

src/providers/ethereum/EthereumMetaMaskProvider.js
```diff
@@ -53,14 +53,22 @@ export default class EthereumMetaMaskProvider extends Provider {
...
+    }
     value = BigNumber(value).toString(16)
 
     const tx = {
       from, to, value, data
     }
 
-    return this._toMM('eth_sendTransaction', tx)
+    const txHash = await this._toMM('eth_sendTransaction', tx)
+    return txHash.replace('0x', '')
```
matthewjablack: Perhaps we should have a `ensureStandardFormat` in utils rather than replacing

examples/browser/bitcoin/ledger-send-transaction.html
monokh: I thought as part of the example, it would be more clear if the ports were the default RPC ports. If you need proxy/ how to setup proxy should be based on individual needs. In my case for example, i actually setup the proxy on `18332` and the bitcoin node is running on a different port.

src/providers/ethereum/EthereumMetaMaskProvider.js
monokh: good idea will add this

matthewjablack (approved):
LGTM 🎉 

`formatEthResponse` should probably use `ensureHexStandardFormat`/`ensureHexEthFormat`

PR #59, opened 2018-09-09
Proper address format

matthewjablack:
### Description

Add proper format for `getAddresses` functions based on [Wallet Provider Specs](https://gist.github.com/harshjv/4136c9af33e8ab1dbbda03d02cf5a6f2#1-walletprovider)

- [x] Proper format for `getAddresses`
- [x] Proper format for `getUsedAddresses`
- [x] Proper format for `getUnusedAddresses`

- [x] `getUsedAddresses` MetaMask provider

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -347,10 +347,10 @@ export default class Client {
    *  of accounts.
    *  Rejects with InvalidProviderResponseError if provider's response is invalid.
    */
-  async getAddresses (config) {
+  async getAddresses () {
     const provider = this.getProviderForMethod('getAddresses')
 
-    const addresses = await provider.getAddresses(config)
+    const addresses = await provider.getAddresses()
 
     if (!isArray(addresses)) {
       throw new InvalidProviderResponseError('Provider returned an invalid response')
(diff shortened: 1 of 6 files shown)


src/Client.js
```diff
@@ -365,10 +365,10 @@ export default class Client {
    *  of accounts.
    *  Rejects with InvalidProviderResponseError if provider's response is invalid.
    */
-  async getUsedAddresses (config) {
+  async getUsedAddresses (startingIndex = 0, numAddresses, config) {
```
monokh: can we remove the config parameter from all the methods? We aren't using them and they aren't part of the spec https://gist.github.com/harshjv/4136c9af33e8ab1dbbda03d02cf5a6f2
the parameter is also very chain specific to be inside the signature for each method. If there's a config option for the provider so that it functions using segwit then that can be something set on the constructor.

src/providers/ethereum/EthereumMetaMaskProvider.js
```diff
@@ -45,6 +45,10 @@ export default class EthereumMetaMaskProvider extends Provider {
     return this._toMM('eth_accounts')
   }
 
+  async getUsedAddresses (startingIndex, numAddresses) {
```
monokh, line 48: what about `getUnusedAddresses` for ethereum?

monokh: Consider if a client has to know that ethereum only has used addresses then we have not normalised

monokh (commented):
I would like to leave this to @harshjv to review on this as he has ongoing work in this area

src/Client.js
matthewjablack: Good idea

src/providers/ethereum/EthereumMetaMaskProvider.js
matthewjablack: So what should it return. Nothing?

monokh: It depends on the context, maybe metamask is able to provide an unused address since it's a HD wallet? Not sure if that is the general expectation when using ethereum though

matthewjablack: Metamask isn't able to provide anything other than the main address. That's ok, it can return nothing for now. We still have the ability to `getUnusedAddresses` for `EthereumLedgerProvider`.

PR #60, opened 2018-09-10
Add Bitcoin ledger provider split transaction segwit setting

matthewjablack:
### Description

When first implementing Ledger Inputs we intended to allow for swaps with non segwit transaction. However we didn't account for the possibility of funding legacy addresses with segwit transactions. Therefore, certain transaction splitting would result in ridiculous values being incorrectly calculated for `amount` and `script`. This PR fixes that. 

- [x] isSegwitSupported: true


monokh (approved):
LGTM

PR #61, opened 2018-09-12
Block checkout swap

matthewjablack:
### Description

This PR adds all the necessary functionality to make CAL work swaps from beginning to detection of Party B transaction confirmation. 

- [x] checkBlockSwap on BTC/ETH Chains
- [x] getAddresses startingIndex and numAddresses BitcoinLedgerProvider
- [x] getBlockHeight on BTC/ETH Chains
- [x] Fix bug with `formatEthResponse` with objects in format `[{},{}]` failing
- [x] Modify `getTransactionByHash` to account for node address information returning error `non wallet id or transaction`

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -342,6 +342,18 @@ export default class Client {
   }
 
   /**
+   * Get network name of current provider
+   * @return {Promise<string, InvalidProviderResponseError>} Resolves with network
+   *  name.
+   *  Rejects with InvalidProviderResponseError if provider's response is invalid.
+   */
+  getNetworkName () {
+    const provider = this.getProviderForMethod('getNetworkName')
+
+    return provider.getNetworkName()
+  }
+
+  /**
    * Get addresses/accounts of the user.
    * @return {Promise<string, InvalidProviderResponseError>} Resolves with a list
    *  of accounts.
@@ -466,6 +478,20 @@ export default class Client {
   }
 
   /**
+   * Decode Transaction from Hex
+   * @param {!string} rawTransaction - A raw transaction usually in the form of a
+   *  hexadecimal string that represents the serialized transaction.
+   * @return {Promise<string, InvalidProviderResponseError>} Resolves with an
+   *  decoded transaction object.
+   *  Rejects with InvalidProviderResponseError if provider's response is invalid.
+   */
+  async decodeRawTransaction (rawTransaction) {
+    const provider = this.getProviderForMethod('decoderawtransaction')
+
+    return provider.decodeRawTransaction(rawTransaction)
+  }
+
+  /**
    * Generate a secret.
    * @param {!string} message - Message to be used for generating secret.
    * @return {Promise<string, null>} Resolves with a secret.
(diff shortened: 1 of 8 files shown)


src/Client.js
```diff
@@ -498,6 +524,18 @@ export default class Client {
     return provider.generateSwap(recipientAddress, refundAddress, secretHash, expiration)
   }
 
+  /**
+   * Generate script hex based on ddress
+   * @param {!string} address - Address for generating Script.
+   * @return {Promise<string, TypeError>} Resolves with redeem swap contract bytecode.
+   *  Rejects with InvalidProviderResponseError if provider's response is invalid.
+   */
+  generateScript (address) {
```
monokh: If this is a chain specific thing, it doesn't need to be here, you can still use `.getMethod()` even without this

src/providers/bitcoin/BitcoinSwapProvider.js
```diff
@@ -66,4 +66,30 @@ export default class BitcoinSwapProvider extends Provider {
...
+    const txs = await Promise.all(transactions.map(async transaction => {
+      return this.getMethod('decodeRawTransaction')(transaction)
+    }))
+    const swapTx = txs
+      .map(transaction => transaction._raw.data)
+      .map(transaction => transaction.vout
+        .map(vout => { return { txid: transaction.txid, scriptPubKey: vout.scriptPubKey } }))
+      .reduce((acc, val) => acc.concat(val), [])
+      .find(txKeys => txKeys.scriptPubKey.hex === sendScript)
+    if (swapTx) {
```
monokh: just a small recommendation, i would write this like: `return swapTx ? swapTx.txid : false`

monokh: also i think this should probably return null instead of false?

monokh: perhaps then this function makes more sense with a name relating to it retrieving transaction for a swap.
`getSwapTransaction(...)` that would return a normalised transaction? this way it would also be multi purpose in a way which you can also tell how many confirmations as well as the transaction id

src/providers/bitcoin/BitcoinSwapProvider.js
```diff
@@ -66,4 +66,30 @@ export default class BitcoinSwapProvider extends Provider {
   refundSwap (pubKey, signature) {
     return this._spendSwap(signature, pubKey, false)
   }
+
+  async checkBlockSwap (blockNumber, recipientAddress, refundAddress, secretHash, expiration) {
+    const data = this.generateSwap(recipientAddress, refundAddress, secretHash, expiration)
+    const scriptPubKey = padHexStart(data)
+    const networkName = this.getMethod('getNetworkName')()
```
monokh: I took a different approach this, instead of having a method on the client, i also made the swap provider accept the network in constructor. I think since it's not required for every chain/provider that is fine

monokh (dismissed):

src/Client.js
matthewjablack: Good point. Removed.

src/providers/bitcoin/BitcoinSwapProvider.js
matthewjablack: I like this approach much more. Updated to specify network in the constructor.

src/providers/bitcoin/BitcoinSwapProvider.js
matthewjablack: I like that name much better. Updated to `getSwapTransaction(...)`

matthewjablack: Updated return statement as well 😄

PR #62, opened 2018-09-14
Swap initiation claim e2e

monokh:
### Description

This PR aims to standardise the swap functionality for initiation and claim as well as providing clear examples

- [x] Switch to sha256 for secrets
- [x] Standardise the swap initiation
- [x] Add consistent interface for claiming swaps
- [x] Implement ethereum claim


monokh:
accidentally checked in a bunch of stuff, trying to revert

monokh:
Removed the accidentally checked in stuff 🙂

src/providers/ethereum/EthereumSwapProvider.js
```diff
@@ -64,20 +65,31 @@ export default class EthereumSwapProvider extends Provider {
...
+  }
+
+  async claimSwap (initiationTxHash, value, recipientAddress, refundAddress, secret, expiration) {
+    const initiationTransaction = await this.getMethod('getTransactionReceipt')(initiationTxHash)
+    const data = this.getRedeemSwapData(secret)
+    return this.getMethod('sendTransaction')(initiationTransaction.contractAddress, 0, data)
+  }
+
+  getRedeemSwapData (secret) {
     return padHexStart(secret, 64)
```
matthewjablack, line 89: What's the reason for requiring padding of 64. Shouldn't it just be 32?

src/providers/bitcoin/BitcoinLedgerProvider.js
```diff
@@ -215,14 +215,6 @@ export default class BitcoinLedgerProvider extends Provider {
   async sendTransaction (to, value, data, from) {
     await this._connectToLedger()
 
-    let receivingAddress
-    if (data) {
```
matthewjablack, line 242: What is the point of the data argument now?

monokh: nothing in bitcoin i guess for now since it's only used for send to P2SH address 😄maybe when claiming is added it will be needed. but it's required on the ethereum side

src/providers/ethereum/EthereumSwapProvider.js
monokh: the length is in characters rather than bytes, so `numBytes = padding / 2`

harshjv:
Merging in favour of #63

Issue #55, opened 2018-08-30
Refrain from using third party api's


gr0kchain:
https://github.com/ConsenSys/chainabstractionlayer/blob/e922f712a99c457ccddf607b597ec3e854a393ea/src/providers/bitcoin/BitcoinLedgerProvider.js#L41

getrawtransaction "txid" ( verbose "blockhash" )

NOTE: By default this function only works for mempool transactions. If the -txindex option is
enabled, it also works for blockchain transactions. If the block which contains the transaction
is known, its hash can be provided even for nodes without -txindex. Note that if a blockhash is
provided, only that block will be searched and if the transaction is in the mempool or other
blocks, or if this node does not have the given block available, the transaction will not be found.
DEPRECATED: for now, it also works for transactions with unspent outputs.

Return the raw transaction data.

If verbose is 'true', returns an Object with information about 'txid'.
If verbose is 'false' or omitted, returns a string that is serialized, hex-encoded data for 'txid'.

Arguments:
1. "txid"      (string, required) The transaction id
2. verbose     (bool, optional, default=false) If false, return a string, otherwise return a json object
3. "blockhash" (string, optional) The block in which to look for the transaction

Result (if verbose is not set or set to false):
"data"      (string) The serialized, hex-encoded data for 'txid'

Result (if verbose is set to true):
{
  "in_active_chain": b, (bool) Whether specified block is in the active chain or not (only present with explicit "blockhash" argument)
  "hex" : "data",       (string) The serialized, hex-encoded data for 'txid'
  "txid" : "id",        (string) The transaction id (same as provided)
  "hash" : "id",        (string) The transaction hash (differs from txid for witness transactions)
  "size" : n,             (numeric) The serialized transaction size
  "vsize" : n,            (numeric) The virtual transaction size (differs from size for witness transactions)
  "version" : n,          (numeric) The version
  "locktime" : ttt,       (numeric) The lock time
  "vin" : [               (array of json objects)
     {
       "txid": "id",    (string) The transaction id
       "vout": n,         (numeric)
       "scriptSig": {     (json object) The script
         "asm": "asm",  (string) asm
         "hex": "hex"   (string) hex
       },
       "sequence": n      (numeric) The script sequence number
       "txinwitness": ["hex", ...] (array of string) hex-encoded witness data (if any)
     }
     ,...
  ],
  "vout" : [              (array of json objects)
     {
       "value" : x.xxx,            (numeric) The value in BTC
       "n" : n,                    (numeric) index
[...]

Issue #54, opened 2018-08-30
Refrain from using third party api's


gr0kchain:
https://github.com/ConsenSys/chainabstractionlayer/blob/e922f712a99c457ccddf607b597ec3e854a393ea/src/providers/bitcoin/BitcoinLedgerProvider.js#L37

This can be replaced with (note this will only return information assuming you've used the importaddress rpc call)

listunspent ( minconf maxconf  ["addresses",...] [include_unsafe] [query_options])

Returns array of unspent transaction outputs
with between minconf and maxconf (inclusive) confirmations.
Optionally filter to only include txouts paid to specified addresses.

Arguments:
1. minconf          (numeric, optional, default=1) The minimum confirmations to filter
2. maxconf          (numeric, optional, default=9999999) The maximum confirmations to filter
3. "addresses"      (string) A json array of bitcoin addresses to filter
    [
      "address"     (string) bitcoin address
      ,...
    ]
4. include_unsafe (bool, optional, default=true) Include outputs that are not safe to spend
                  See description of "safe" attribute below.
5. query_options    (json, optional) JSON with query options
    {
      "minimumAmount"    (numeric or string, default=0) Minimum value of each UTXO in BTC
      "maximumAmount"    (numeric or string, default=unlimited) Maximum value of each UTXO in BTC
      "maximumCount"     (numeric or string, default=unlimited) Maximum number of UTXOs
      "minimumSumAmount" (numeric or string, default=unlimited) Minimum sum value of all UTXOs in BTC
    }

Result
[                   (array of json object)
  {
    "txid" : "txid",          (string) the transaction id
    "vout" : n,               (numeric) the vout value
    "address" : "address",    (string) the bitcoin address
    "account" : "account",    (string) DEPRECATED. The associated account, or "" for the default account
    "scriptPubKey" : "key",   (string) the script key
    "amount" : x.xxx,         (numeric) the transaction output amount in BTC
    "confirmations" : n,      (numeric) The number of confirmations
    "redeemScript" : n        (string) The redeemScript if scriptPubKey is P2SH
    "spendable" : xxx,        (bool) Whether we have the private keys to spend this output
    "solvable" : xxx,         (bool) Whether we know how to spend this output, ignoring the lack of keys
    "safe" : xxx              (bool) Whether this output is considered safe to spend. Unconfirmed transactions
                              from outside keys and unconfirmed replacement transactions are considered unsafe
                              and are not eligible for spending by fundrawtransaction and sendtoaddress.
  }
[...]

PR #65, opened 2018-09-27
Add Bitcoin RPC provider

### LINKED ISSUE #53: Refrain from using third party api's
https://github.com/ConsenSys/chainabstractionlayer/blob/e922f712a99c457ccddf607b597ec3e854a393ea/src/providers/bitcoin/BitcoinLedgerProvider.js#L33

Replace this with the rpc's calls to bitcoind using

importaddress "address" ( "label" rescan p2sh )

Adds a script (in hex) or address that can be watched as if it were in your wallet but cannot be used to spend. Requires a new wallet backup.

Arguments:
1. "script"           (string, required) The hex-encoded script (or address)
2. "label"            (string, optional, default="") An optional label
3. rescan               (boolean, optional, default=true) Rescan the wallet for transactions
4. p2sh                 (boolean, optional, default=false) Add the P2SH version of the script as well

Note: This call can take minutes to complete if rescan is true, during that time, other rpc calls
may report that the imported address exists but related transactions are still missing, leading to temporarily incorrect/bogus balances and unspent outputs until rescan completes.
If you have the full public key, you should call importpubkey instead of this.

Note: If you import a non-standard raw script in hex form, outputs sending to it will be treated
as change, and not show up in many RPCs.

getreceivedbyaddress "address" ( minconf )

Returns the total amount received by the given address in transactions with at least minconf confirmations.

Arguments:
1. "address"         (string, required) The bitcoin address for transactions.
2. minconf             (numeric, optional, default=1) Only include transactions confirmed at least this many times.

Result:
amount   (numeric) The total amount in BTC received at this address.

harshjv:
This PR removes API/RPC methods from Hardware provider and simplifies Bitcoin transaction signing. Fixes #53, #54, #55 

- [x] Add `Address`
- [x] Remove blockchain.info API calls from `BitcoinLedgerProvider`
- [x] Introduce `isUsedAddress` method on `Client`
- [x] Add `LedgerProvider`
- [x] Rebase with `master`
- [x] Update docs


src/Client.js
```diff
@@ -359,16 +346,26 @@ export default class Client {
...
+  /**
+   * Get addresses/accounts of the user.
+   * @return {Promise<string, InvalidProviderResponseError>} Resolves with a list
+   *  of accounts.
+   *  Rejects with InvalidProviderResponseError if provider's response is invalid.
+   */
+  async isUsedAddress (address) {
+    const isUsed = await this.getMethod('isUsedAddress')(address)
+
+    return isUsed
```
matthewjablack: Can simplify this to 

`return this.getMethod('isUsedAddress')(address)`

src/providers/LedgerProvider.js
```diff
@@ -0,0 +1,70 @@
+import Provider from '../Provider'
+
+import Transport from '@ledgerhq/hw-transport-node-hid'
+
+export default class LedgerProvider extends Provider {
```
matthewjablack, line 5: Nice, I like it

PR #71, opened 2018-10-04
Find secret

matthewjablack:
### Description

This PR adds  `getSecret` for both Ethereum and Bitcoin swap providers based on claim transaction hash

- [x] `getSecret` for both Ethereum and Bitcoin swap providers

### DIFF
--- a/src/providers/bitcoin/BitcoinSwapProvider.js
+++ b/src/providers/bitcoin/BitcoinSwapProvider.js
@@ -161,6 +161,15 @@ export default class BitcoinSwapProvider extends Provider {
     return swapTx ? swapTx.hash : null
   }
 
+  async getSecret (claimTxHash) {
+    const claimTxRaw = await this.getMethod('getRawTransactionByHash')(claimTxHash)
+    const claimTx = await this.getMethod('decodeRawTransaction')(claimTxRaw)
+    const script = Buffer.from(claimTx._raw.data.vin[0].scriptSig.hex, 'hex')
+    const sigLength = script[0]
+    const secretLength = script.slice(sigLength + 1)[0]
+    return script.slice(sigLength + 2, sigLength + secretLength + 2).toString('hex')
+  }
+
   generateSigTxInput (txHashLE, voutIndex, script) {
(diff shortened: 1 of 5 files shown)


src/providers/bitcoin/BitcoinSwapProvider.js
```diff
@@ -161,6 +161,15 @@ export default class BitcoinSwapProvider extends Provider {
     return swapTx ? swapTx.hash : null
   }
 
+  async getSecret (claimTxHash) {
+    const claimTxRaw = await this.getMethod('getRawTransactionByHash')(claimTxHash)
+    const claimTx = await this.getMethod('splitTransaction')(claimTxRaw, true)
```
monokh: you could also just use rpc `decoderawtransaction`

matthewjablack: Good point, there shouldn't be dependency on `BitcoinLedgerProvider` for something like this.

matthewjablack: Updated to use `BitcoinRPCProvider`

src/providers/ethereum/EthereumMetaMaskProvider.js
```diff
@@ -56,12 +56,13 @@ export default class EthereumMetaMaskProvider extends Provider {
     return addresses[0]
   }
 
-  async signMessage (message, from) {
-    from = String(from)
-
+  async signMessage (message) {
```
harshjv, line 59: This makes sense for MM. Nice.

harshjv: However, if addresses are being managed by the UI, can we assume that UI will send correct address with `signMessage`?

harshjv (approved):
Small comments. Otherwise, LGTM

monokh:
In the interest of not having too many branching going on and since you are away @mattBlackDesign, I'm going to merge this so I can follow up with my work cleanly. Feel free to come back to this after you are back :)

PR #74, opened 2018-10-05
E2E claiming example with chain scanning functions 

monokh:
### Description

Revises the swap transaction finding methods slightly to make them entirely self contained. 

Demonstrates the new chain scanning functions in the e2e example

- [ ] Refactor some of the transaction finding functions to contain all the required logic
- [ ] Added the transaction finding and secret reveal into the e2e example
- [ ] Tests seem to be broken


src/providers/bitcoin/BitcoinSwapProvider.js
```diff
@@ -127,41 +129,64 @@ export default class BitcoinSwapProvider extends Provider {
     return this._spendSwap(signature, pubKey, false)
   }
 
-  async getSwapTransaction (blockNumber, recipientAddress, refundAddress, secretHash, expiration) {
+  async findInitiateSwapTransaction (recipientAddress, refundAddress, secretHash, expiration) {
```
matthewjablack, line 132: I worry about removing `blockNumber` from fn arguments. How can we be sure that the blockHeight is the right place to start scanning the chain. What if the user has been disconnected, and needs to rescan the chain from a certain block (which should be from the spot where party A initiated the Swap in the first place).

matthewjablack: I think we still need blockNumber here, or maybe split this into two functions, one with a start blockNumber to start scanning the chain, and the other with the simple find function for looking at a specific transaction to try and find whether it's a swap transaction.

monokh: my assumption for this function was that it's like subscribing for an event. you would start listening the moment you were waiting for the transaction i.e. the flow would be:
```
Party A: initiateSwap(), findInitiateSwapTransaction(partyB)
Party B: initiateSwap(), findClaimSwapTransaction(partyA)
Party A: findInitiateSwapTransaction is resolved, claimSwap(), findClaimSwapTransaction(partyB)
PartyB: findClaimSwapTransaction is resolved, claimSwap(), swap done
PartyA: findClaimSwapTransaction is resolved, swap done
```
There is the concern that the user has been disconnected but I couldn't really see how we were already solving this problem. Scanning for a transaction under a given block number would allow for this possibility but we haven't integrated it. It feels like the approach would be too many requests for mainnet for any chain that has frequent blocks and more than a few transactions for a given block
Was there a plan/ideas for integrating the disconnectivity problem?

monokh: i guess the assumption that this implementation makes is that the user would be online for the events to occur. can you clarify if we had plans/had already working recovering from a disconnection when scanning the chain?

monokh: Discussed with Matt that we can keep the previous function intact as it would allow recovering from disconnecting when/if we wanted to implement it. Also discussed probably this could be something provided by another function in the swap provider too.

monokh: So i tried to keep the function and use it internally in `findInitiateSwapTransaction` and it became horribly inefficient because it was doing a bunch of stuff it only needed to do once over and over again for the sake of reusability. 

Due to this, I would say we can leave it out and then introduce recovering from a disconnection when we are looking to add it. It would be a reuse of the current function.

matthewjablack: This makes sense. We can look into reintroducing it later in the roadmap 👍

matthewjablack (approved):
With new changes LGTM 🎉

PR #27, opened 2018-08-02
Fix circleci docker image

matthewjablack:
This PR bumps `chainabstractionlayer-dev` docker version to 0.0.5 which includes `python2.7` which is required for `secp256k1` and `git-core` which is required for npm `usb` which is needed for `@ledgerhq/hw-transport-node-hid`. 

It also changes the usage of npm `usb` to use fork [https://github.com/mattBlackDesign/node-usb](https://github.com/mattBlackDesign/node-usb) which is a temporary fix. This fork removes functionality of usb, in order for it to work properly in circleci (Docker doesn't have access to host usb ).

PR #78, opened 2018-10-09
Update hw-transports

harshjv:
### Description

- [x] Update ledger-hw transports

See https://github.com/liquality/chainabstractionlayer/pull/27#issuecomment-410671464 for the reference.


monokh (approved):
LGTM :)

PR #79, opened 2018-10-10
Fix isAddressUsed BTC

matthewjablack:
### Description

This PR fixes isAddressUsed for BTC, by using `receivedbyaddress` instead of `getunspenttransactions`

- [ ] Add `receivedByAddress`

### DIFF
--- a/src/providers/bitcoin/BitcoinRPCProvider.js
+++ b/src/providers/bitcoin/BitcoinRPCProvider.js
@@ -14,7 +14,7 @@ export default class BitcoinRPCProvider extends JsonRpcProvider {
 
   async isAddressUsed (address) {
     address = String(address)
-    const utxo = await this.getUnspentTransactions(address)
+    const utxo = await this.getReceivedByAddress(address)
 
     return utxo.length !== 0
   }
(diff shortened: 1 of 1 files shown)


src/providers/bitcoin/BitcoinRPCProvider.js
```diff
@@ -14,7 +14,7 @@ export default class BitcoinRPCProvider extends JsonRpcProvider {
 
   async isAddressUsed (address) {
     address = String(address)
-    const utxo = await this.getUnspentTransactions(address)
+    const utxo = await this.getReceivedByAddress(address)
```
monokh: something about this doesn't seem clear enough...
I'm looking at the return of `getreceivedbyaddress` RPC call, and it is the amount received by this address but to check this we do `utxo.length` as if the return would be a list of utxos. wouldn't the check be more like this?
```
const amoutnReceived = await this.getReceivedByAddress(address)
return amountReceived > 0
```

monokh (dismissed):

src/providers/bitcoin/BitcoinRPCProvider.js
matthewjablack: @monokh good catch. Updated with proper logic.

PR #80, opened 2018-10-10
E2e claiming with chain scanning

monokh:
### Description

Fixes issue with bitcoin claiming where it wasn't using the correct path under ledger.


monokh:
Missed pushing this commit on PR #74

PR #82, opened 2018-10-11
Swap providers: finding transaction ensures correct value locked

monokh:
### Description

Bug: the finding of the initiate transaction does not check the value locked using the transaction. It would open potential attack vectors. 

- [x] Check the value locked in the swap when finding/verifying an initiation transaction

### DIFF
--- a/examples/swap/swap-e2e.html
+++ b/examples/swap/swap-e2e.html
@@ -94,52 +94,40 @@
     }
 
     function initiateSwap (client, prefix) {
-      client.createSwapScript(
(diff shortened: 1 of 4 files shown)


harshjv (approved):
LGTM.

Just a thought regarding adding new parameters to a function. Should there be a precedence over the order in which they appear?

monokh:
@harshjv yea i just made up the order initially, but have stuck to it as much as i could among the swap functions. They should be largely consistent unless i've missed something

PR #83, opened 2018-10-12
Add verify swap method

monokh:
### Description

Adds the verify swap method and makes sure find transaction methods return a full transaction object.

- [x] Bitcoin swap provider find functions return full tx object
- [x] Add verify swap method
- [x] Simplify a little looking for transactions in a block
- [x] Add confirmations to the standard tx object
- [x] Ensure eth transaction objects are normalized
- [x] Standardise eth addresses for the swap provider

### DIFF
--- a/src/providers/bitcoin/BitcoinSwapProvider.js
+++ b/src/providers/bitcoin/BitcoinSwapProvider.js
@@ -145,24 +145,15 @@ export default class BitcoinSwapProvider extends Provider {
         block = await this.getMethod('getBlockByNumber')(blockNumber, true)
       } catch (e) { }
       if (block) {
-        const transactions = block.transactions
-        const txs = await Promise.all(transactions.map(async transaction => {
-          return this.getMethod('decodeRawTransaction')(transaction)
-        }))
-        initiateSwapTransaction = txs
-          .map(transaction => transaction._raw.data)
-          .map(transaction => transaction.vout
-            .map(vout => { return { txid: transaction.txid, scriptPubKey: vout.scriptPubKey, value: vout.valueSat } }))
-          .reduce((acc, val) => acc.concat(val), [])
-          .find(txInfo => txInfo.scriptPubKey.hex === sendScript && txInfo.value === value)
+        initiateSwapTransaction = block.transactions.find(tx =>
+          tx._raw.data.vout.find(vout => vout.scriptPubKey.hex === sendScript && vout.valueSat === value)
+        )
         blockNumber++
       }
       await sleep(5000)
     }
 
-    return {
-      hash: initiateSwapTransaction.txid // TODO: full transaction object
-    }
+    return initiateSwapTransaction
   }
 
   async findClaimSwapTransaction (initiationTxHash, secretHash) {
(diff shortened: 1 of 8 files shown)


src/providers/bitcoin/BitcoinSwapProvider.js
```diff
@@ -145,24 +153,15 @@ export default class BitcoinSwapProvider extends Provider {
...
-          .map(transaction => transaction.vout
-            .map(vout => { return { txid: transaction.txid, scriptPubKey: vout.scriptPubKey, value: vout.valueSat } }))
-          .reduce((acc, val) => acc.concat(val), [])
-          .find(txInfo => txInfo.scriptPubKey.hex === sendScript && txInfo.value === value)
+        initiateSwapTransaction = block.transactions.find(tx =>
+          this.doesTransactionMatchSwapParams(tx, value, recipientAddress, refundAddress, secretHash, expiration)
+        )
         blockNumber++
       }
       await sleep(5000)
```
matthewjablack, line 161: Just noticed `sleep(5000)`

This means a 5 second delay correct? Should we reduce this to enable faster block scanning?

src/providers/ethereum/EthereumSwapProvider.js
```diff
@@ -95,15 +95,24 @@ export default class EthereumSwapProvider extends Provider {
     return ''
   }
 
-  async findInitiateSwapTransaction (value, recipientAddress, refundAddress, secretHash, expiration) {
+  doesTransactionMatchSwapParams (transaction, value, recipientAddress, refundAddress, secretHash, expiration) {
     const data = this.createSwapScript(recipientAddress, refundAddress, secretHash, expiration)
+    return transaction.input === data.toLowerCase() && transaction.value === value
```
matthewjablack: Should we standardize `.toLowerCase()` throughout Ethereum providers (seems to be an issue that comes up a lot, and we can't lowerCase on bitcoin side due to base58). Maybe future PR.

matthewjablack (approved):
Just a few comments that can be done in future PR's. Other than that LGTM 🎉

src/providers/bitcoin/BitcoinSwapProvider.js
monokh: Yes 5s. I'm not sure what to exactly do with the wait times because it depends on quite a few factors. The minimum blocktime on ethereum is for example 15s atm. I just picked 5s randomly 🤷‍♂️ 
I think eventually we need to think of this properly, like events on blocks being available, perhaps an option on the provider for how often to scan the chain etc.

src/providers/ethereum/EthereumSwapProvider.js
monokh: Good point, added this as a `ensureAddressStandardFormat`

src/providers/bitcoin/BitcoinSwapProvider.js
matthewjablack: Agreed @monokh  I think we also need to think about how fast we want to scan the chain in certain instances. For example if we're significantly behind, then it would make sense to scan asynchronously as fast as possible, but if the current block is latest block then a longer delay makes sense.

matthewjablack: Maybe it would makes sense to check if it's the latest block, and if so, add `sleep` in there

monokh: Wouldn't you then be spamming the node? 😕  feel a little uncomfortable about something like that atm

matthewjablack: Yep we'd have to check how much this will affect the performance of the node.

PR #85, opened 2018-10-17
Add logic for btc refund

matthewjablack:
### Description

This PR adds logic for BTC Swap Refund

- [x] add `refundSwap` fn
- [x] add `_redeemSwap` for use with both `claimSwap` and `refundSwap`
- [x] fix issue with `getDerivationPathFromAddress`
- [x] add refund button to `swap-e2e.html`
- [x] add scriptNumEncode for encoding numbers
- [x] modify locktime and sequence for CLTV

### DIFF
--- a/examples/swap/swap-e2e.html
+++ b/examples/swap/swap-e2e.html
@@ -44,11 +44,12 @@
           <img style="display:none" id="btcVerifySwapTrue" src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABgAAAAYCAYAAADgdz34AAACHElEQVRIS+2VPU/bQBjH/5dA6teEDpWqdmThG3RoJip1Z2eq8hm68rL3G9BurHyASnSCod+gC0srkJCIII7t2Djxof/hC+fYJFERG5YenXM6/37PPc/ZEXjmSzwzHy8CVeEvPz++a6yI83ws3//4fHphlv3JJdLw9Y11nP05w6zkSQITnmYpkizB1d+rkuS/BXVwXRpKDjZPFLsimFdPDTDhyW2iMhfiHkU4JPYOPp3sVgSL6mk2lDUfpSMkaQI0oAT9f/0SvCTQ8LW3a7AtG47lVJpmJhCPYsRJrMCiIXB9cV2BlwS9X11JuJQSMpdwbAeu404lXMyjyMyjKEIURwrMGFwOauFlwXF3FwI77Tdt5HmuJJ7rwfM8JeFFeDgMMYyG08zDfvgovNKDXiFxX7tKwPB9H37bV4IgCDAMHuDxIJ4Lrz1FWmL51lTS6XQgIRHcBOrcsSxplC6E1wo4qSUtu3UvkXL69rOpWZotBX9UYEqaq03ClITwyXiyNHyuwJSwJFpivkTL/Jcs+lSI3nF3h6eLMAm5/33zdG8GrOv3UEdjgSngfQPACoDVYmxybvvow1cpIA63fn8zniWQkRcxATAGkBUj56UpILwFwAPAc2kVIkqUqPh28RkTTjCD4BGAsIhbimcFzN4G4AB4ZezCFOhN6Oy1gNmnAOJCxN8lgW46M9Why8axrl/mTnhPmRarnixq8jIHZe6aO75QNCiLJ0zzAAAAAElFTkSuQmCC" />
           <img style="display:none" id="btcVerifySwapFalse" src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABgAAAAYCAYAAADgdz34AAACT0lEQVRIS8WVS2sTURTHf2N8JMG2iK/W4luhKkpRSqmvb6BfoNCVazcKbsRUN4Li1oVLwS9Qv4GvEKqlKNWitmqTaavNQ6skrUlz5Qxzp8nMnWk3wYHLZOae8/+dxz0TixZfVov1+b8Am+SlbspPo7JcyyY0A3G0YETB5TDIemyMAO3Ydf4scy9eYYJ4NkODzD1+YrSRzAMA7dh5bsCrzPzLdJNAozilEqpYwm+jnQOAWZKqc6A/UPb5dMaBuFGNdA0NQrGIKhRXA0ln2EO5SdOUwbAFqd39fQHI98yo807EVaGAyhc8G9lTcLub8nCjY1gPHMiuvjMBSOxYDyqfp76Q9/Z+jL4xiht7oL1skg5k5+neyFlcGBsPFY8EyKaG7Og9ZYTkx99Giq8JEANp+vaTJ4yAwruJQFP9hpGfChGXhi6/HgstU/H9h0hIKECL1+xZarbtAJRpcIDS5MdQiBGgxas5m2o250X/69Nn53fH0SOBjGTPPwPGHnji2RzVmeyq+NS001DXKdVx+FAQMjW9/kFrP3jAE1j88rXptOjTFWWjnSMHrW3/Pn5/m2EF7uyl7ESvryzJVAxuaRvTFPtLJLANwEZg0ySJm+1YN36i7h+n8sDtb2NAaoLEtW1Y1xdR93qo3AVqQNW91+VcNDqI+GZgK9AGxJ8Rv3qRpUdAzIWLvSw5ULLqz4lfucDSQ1e4Avxx11/Z9wMk+gSQBLa42Yh4I0BXyQGAVNBZEv0yUAYEJM9NAF0yyUQvXTYduf/keJm4GQlIg+Xe+j/9fwC4AyiJR4XmAAAAAElFTkSuQmCC" />
         </p>
-        <h3>Claim swap:</h3>
+        <h3>Redeem swap:</h3>
         <p><strong>Wait until initiation transaction has been confirmed!</strong></p>
         <p>secret: <input type="text" id="btcClaimSecret" /></p>
         <p>initiationTxHash: <input type="text" id="btcInitiationTxHash" /></p>
         <p><button id="btcClaimSwap">Claim Swap</button></p>
+        <p><button id="btcRefundSwap">Refund Swap</button></p>
         <p>transaction hash: <textarea id="btcClaimSwapResult"></textarea></p>
         <p style="color: red; display: none;" id="btcFindClaimSwap">Scanning for claim tx: <textarea id="btcFindClaimSwapTransactionResult"></textarea></p>
         <p style="color: red; display: none;" id="btcFindClaimSwapSecret">Secret Revealed: <textarea id="btcFindClaimSwapTransactionResultSecret"></textarea></p>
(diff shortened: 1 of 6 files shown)


src/providers/LedgerProvider.js
```diff
@@ -44,7 +44,7 @@ export default class LedgerProvider extends Provider {
 
     while (!path) {
       const addr = await this.getAddresses(index, 1)
-      if (String(addr) === address) path = this.getDerivationPathFromIndex(index)
+      if (String(addr[0].address) === address) path = this.getDerivationPathFromIndex(index)
```
harshjv: You can just use `String(addr[0])`. All `Address` objects have `.toString()` methods.

examples/swap/swap-e2e.html
```diff
@@ -44,11 +44,12 @@ <h3>Verify swap:</h3>
...
           <img style="display:none" id="btcVerifySwapFalse" src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABgAAAAYCAYAAADgdz34AAACT0lEQVRIS8WVS2sTURTHf2N8JMG2iK/W4luhKkpRSqmvb6BfoNCVazcKbsRUN4Li1oVLwS9Qv4GvEKqlKNWitmqTaavNQ6skrUlz5Qxzp8nMnWk3wYHLZOae8/+dxz0TixZfVov1+b8Am+SlbspPo7JcyyY0A3G0YETB5TDIemyMAO3Ydf4scy9eYYJ4NkODzD1+YrSRzAMA7dh5bsCrzPzLdJNAozilEqpYwm+jnQOAWZKqc6A/UPb5dMaBuFGNdA0NQrGIKhRXA0ln2EO5SdOUwbAFqd39fQHI98yo807EVaGAyhc8G9lTcLub8nCjY1gPHMiuvjMBSOxYDyqfp76Q9/Z+jL4xiht7oL1skg5k5+neyFlcGBsPFY8EyKaG7Og9ZYTkx99Giq8JEANp+vaTJ4yAwruJQFP9hpGfChGXhi6/HgstU/H9h0hIKECL1+xZarbtAJRpcIDS5MdQiBGgxas5m2o250X/69Nn53fH0SOBjGTPPwPGHnji2RzVmeyq+NS001DXKdVx+FAQMjW9/kFrP3jAE1j88rXptOjTFWWjnSMHrW3/Pn5/m2EF7uyl7ESvryzJVAxuaRvTFPtLJLANwEZg0ySJm+1YN36i7h+n8sDtb2NAaoLEtW1Y1xdR93qo3AVqQNW91+VcNDqI+GZgK9AGxJ8Rv3qRpUdAzIWLvSw5ULLqz4lfucDSQ1e4Avxx11/Z9wMk+gSQBLa42Yh4I0BXyQGAVNBZEv0yUAYEJM9NAF0yyUQvXTYduf/keJm4GQlIg+Xe+j/9fwC4AyiJR4XmAAAAAElFTkSuQmCC" />
         </p>
-        <h3>Claim swap:</h3>
+        <h3>Redeem swap:</h3>
```
monokh, line 47: I'd keep this the same as the method name for clarity

monokh: can think of revamping the name of the methods though if you want to have a discussion around that

examples/swap/swap-e2e.html
```diff
@@ -95,12 +96,12 @@ <h3>Claim swap:</h3>
     /* global $, ChainAbstractionLayer, web3 */
     const { Client, providers, networks, crypto } = ChainAbstractionLayer
     const bitcoin = new Client()
-    bitcoin.addProvider(new providers.bitcoin.BitcoinRPCProvider('https://localhost:18332', 'bitcoin', 'local321'))
-    bitcoin.addProvider(new providers.bitcoin.BitcoinLedgerProvider({ network: networks.bitcoin_testnet }))
+    bitcoin.addProvider(new providers.bitcoin.BitcoinRPCProvider('http://localhost:8000', 'bitcoin', 'local321'))
+    bitcoin.addProvider(new providers.bitcoin.BitcoinLedgerProvider({ network: networks.bitcoin_testnet, segwit: false }))
```
monokh, line 110: if you made `segwit` default to false then won't need to specify it

examples/swap/swap-e2e.html
```diff
@@ -95,12 +96,12 @@ <h3>Claim swap:</h3>
     /* global $, ChainAbstractionLayer, web3 */
     const { Client, providers, networks, crypto } = ChainAbstractionLayer
     const bitcoin = new Client()
-    bitcoin.addProvider(new providers.bitcoin.BitcoinRPCProvider('https://localhost:18332', 'bitcoin', 'local321'))
-    bitcoin.addProvider(new providers.bitcoin.BitcoinLedgerProvider({ network: networks.bitcoin_testnet }))
+    bitcoin.addProvider(new providers.bitcoin.BitcoinRPCProvider('http://localhost:8000', 'bitcoin', 'local321'))
```
monokh: you don't have issues running without https? 🤔 
Can't remember for what reason but I had to make it run over https, might have been because of ledger, having to run the example through https

i'd keep the default bitcoin port in the example that's committed to master

src/Client.js
```diff
@@ -547,29 +547,21 @@ export default class Client {
...
-      throw new TypeError('PubKey should be a valid hex string')
-    }
-
-    if (!isString(signature)) {
-      throw new TypeError('Signature should be a string')
-    }
-
-    if (!(/^[A-Fa-f0-9]*$/.test(signature))) {
-      throw new TypeError('Signature should be a valid hex string')
+  async refundSwap (initiationTxHash, recipientAddress, refundAddress, secret, expiration) {
```
monokh: finding it odd that secret is required for refund. Party B wouldn't have the secret right? only the secret hash

examples/swap/swap-e2e.html
matthewjablack: The problem is we made the entire object `{ network: networks.bitcoin_testnet, segwit: false }` to a certain default. So if you don't specify `segwit` here it actually comes up as `undefined`.

examples/swap/swap-e2e.html
matthewjablack: I think we need to define what the words `claim`, `redeem` and `refund` really mean. I think `redeem` applies to the redeeming of a swap contract whether it be `claim` or `refund`. `claim ` and `refund` on the other hand, are only for those specific use cases. (It's kind of like redeeming a p2sh can be a `claim` or `refund`)

matthewjablack: In this case, I think `Redeem` makes the most sense, because it applies to both `claim` and `refund`.

examples/swap/swap-e2e.html
matthewjablack: Maybe we should actually have a config file or ENV vars for this

src/providers/LedgerProvider.js
matthewjablack: Good catch. Reverted

examples/swap/swap-e2e.html
matthewjablack: Reverted for now

src/Client.js
matthewjablack: Yep very good point. Changed to use secretHash instead.

examples/swap/swap-e2e.html
monokh: ah yes good point there

monokh: ah yes good point. Although you could also turn this into regular arguments instead of an object and it would work that way
Not sure if you had some intentions with using an object here

monokh: ah yes good point. Although you could also turn this into regular arguments instead of an object and it would work that way
Not sure if you had some intentions with using an object here

monokh: ah yes good point. Although you could also turn this into regular arguments instead of an object and it would work that way
Not sure if you had some intentions with using an object here

src/providers/bitcoin/BitcoinSwapProvider.js
```diff
@@ -52,7 +50,16 @@ export default class BitcoinSwapProvider extends Provider {
...
+    return this._redeemSwap(initiationTxHash, recipientAddress, refundAddress, secret, expiration, true)
+  }
+
+  async refundSwap (initiationTxHash, recipientAddress, refundAddress, secretHash, expiration) {
+    return this._redeemSwap(initiationTxHash, recipientAddress, refundAddress, secretHash, expiration, false)
+  }
+
+  async _redeemSwap (initiationTxHash, recipientAddress, refundAddress, secretParam, expiration, isClaim) {
+    const secretHash = isClaim ? sha256(secretParam) : secretParam
+    const lockTime = isClaim ? 0 : expiration + 100
```
monokh, line 61: why the +100?

src/providers/bitcoin/BitcoinSwapProvider.js
```diff
@@ -15,10 +15,8 @@ export default class BitcoinSwapProvider extends Provider {
   }
 
   createSwapScript (recipientAddress, refundAddress, secretHash, expiration) {
-    let expirationHex = Buffer.from(padHexStart(expiration.toString(16)), 'hex').reverse()
-
-    expirationHex = expirationHex.slice(0, Math.min(expirationHex.length, 5))
-    expirationHex.writeUInt8(0, expirationHex.length - 1)
+    // let expirationHex = Buffer.from(padHexStart(expiration.toString(16)), 'hex').reverse()
```
monokh: is the comment needed?

src/providers/bitcoin/BitcoinSwapProvider.js
```diff
@@ -52,7 +50,16 @@ export default class BitcoinSwapProvider extends Provider {
...
   async claimSwap (initiationTxHash, recipientAddress, refundAddress, secret, expiration) {
-    const secretHash = sha256(secret)
+    return this._redeemSwap(initiationTxHash, recipientAddress, refundAddress, secret, expiration, true)
+  }
+
+  async refundSwap (initiationTxHash, recipientAddress, refundAddress, secretHash, expiration) {
+    return this._redeemSwap(initiationTxHash, recipientAddress, refundAddress, secretHash, expiration, false)
+  }
+
+  async _redeemSwap (initiationTxHash, recipientAddress, refundAddress, secretParam, expiration, isClaim) {
```
monokh, line 59: I have to say the structure of this method has become odd now. It feels like you are not making the right abstractions

For example `isClaim ? recipientAddress : refundAddress` is repeated multiple times. An alternative would be reworking this method to only accept a `to` and pass the result of `isClaim ? recipientAddress : refundAddress` to it instead. 

Similar story for the other params based on `isClaim`

src/providers/bitcoin/BitcoinUtil.js
```diff
@@ -69,10 +69,41 @@ function reverseBuffer (src) {
...
+function scriptNumSize (i) {
+  return i > 0x7fffffff ? 5
+  : i > 0x7fffff ? 4
+  : i > 0x7fff ? 3
+  : i > 0x7f ? 2
+  : i > 0x00 ? 1
+  : 0
+}
+
+function scriptNumEncode (number) {
```
monokh, line 81: Little bit uncomfortable with writing something like this ourselves, could we have used a lib? could do with a few test cases otherwise

monokh (commented):
Left some comments around things that can be improved. If it works, i don't want to stop it going forward but if you think one of the improvements is necessary, go ahead with that

src/providers/bitcoin/BitcoinUtil.js
matthewjablack: This is copied from bitcoinjslib so should be safe

src/providers/bitcoin/BitcoinSwapProvider.js
matthewjablack: Because locktime must be greater than locktime set in the redeem script in order to pass.

src/providers/bitcoin/BitcoinSwapProvider.js
matthewjablack: Nope, removed

src/providers/bitcoin/BitcoinSwapProvider.js
matthewjablack: I agree. Modified to be more abstract and reuse conditional.

src/providers/bitcoin/BitcoinSwapProvider.js
matthewjablack: I chose 100 arbitrarily (100 seconds seems reasonable).

PR #86, opened 2018-10-18
EthereumSwapProvider: refund

monokh:
### Description

Implement refund method for ethereum swap

### DIFF
--- a/src/providers/ethereum/EthereumSwapProvider.js
+++ b/src/providers/ethereum/EthereumSwapProvider.js
@@ -83,16 +83,13 @@ export default class EthereumSwapProvider extends Provider {
 
   async claimSwap (initiationTxHash, recipientAddress, refundAddress, secret, expiration) {
     const initiationTransaction = await this.getMethod('getTransactionReceipt')(initiationTxHash)
-    const data = this.getRedeemSwapData(secret)
+    const data = padHexStart(secret, 64)
     return this.getMethod('sendTransaction')(initiationTransaction.contractAddress, 0, data)
   }
 
-  getRedeemSwapData (secret) {
-    return padHexStart(secret, 64)
-  }
-
-  getRefundSwapData () {
-    return ''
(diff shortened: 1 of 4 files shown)


src/providers/ethereum/EthereumSwapProvider.js
```diff
@@ -83,16 +83,13 @@ export default class EthereumSwapProvider extends Provider {
 
   async claimSwap (initiationTxHash, recipientAddress, refundAddress, secret, expiration) {
     const initiationTransaction = await this.getMethod('getTransactionReceipt')(initiationTxHash)
-    const data = this.getRedeemSwapData(secret)
+    const data = padHexStart(secret, 64)
     return this.getMethod('sendTransaction')(initiationTransaction.contractAddress, 0, data)
   }
 
-  getRedeemSwapData (secret) {
```
monokh, line 90: found these methods pretty unnecessary

matthewjablack: yep, good call removing them

src/providers/ethereum/EthereumSwapProvider.js
```diff
@@ -83,16 +83,13 @@ export default class EthereumSwapProvider extends Provider {
...
-    return padHexStart(secret, 64)
-  }
-
-  getRefundSwapData () {
-    return ''
+  async refundSwap (initiationTxHash, recipientAddress, refundAddress, secretHash, expiration) {
```
matthewjablack, line 90: Should this be standardized as `secretHash` as an argument or `secret` as an argument?

matthewjablack (approved):
LGTM 🎉  just one small comment

src/providers/ethereum/EthereumSwapProvider.js
monokh: secrethash i believe as this is what the contract is created with. If secret was part of the interface, a party not knowing the secret couldn't get a refund which doesn't seem right

PR #88, opened 2018-10-23
Fix ethereum isAddressUsed

monokh:
### Description

Fixes the isAddressUsed function for ethereum. Normalises the address and makes sure required rpc params are there


matthewjablack (approved):
LGTM 🎉

PR #91, opened 2018-11-09
Ledger provider improvements

monokh:
### Description

Makes improvements to the ledger provider

- [x] Set limit for ledger lookup by address
- [x] Add cache for addresses from ledger
- [x] Improve logic for determining utxos - get every single address upto a max sometimes is not enough and in other cases can be lookups that are not necessary

### DIFF
--- a/src/providers/LedgerProvider.js
+++ b/src/providers/LedgerProvider.js
@@ -12,6 +12,7 @@ export default class LedgerProvider extends Provider {
 
     this._App = App
     this._baseDerivationPath = baseDerivationPath
+    this._addressCache = {}
   }
 
   async createTransport () {
@@ -61,7 +62,12 @@ export default class LedgerProvider extends Provider {
 
   async getAddressFromIndex (addressIndex) {
     const path = this.getDerivationPathFromIndex(addressIndex)
-    return this.getAddressFromDerivationPath(path)
+    if (path in this._addressCache) {
+      return this._addressCache[path]
+    }
+    const address = await this.getAddressFromDerivationPath(path)
+    this._addressCache[path] = address
+    return address
   }
 
   async getAddresses (startingIndex = 0, numAddresses = 1) {
--- a/src/providers/LedgerProvider.js
+++ b/src/providers/LedgerProvider.js
@@ -42,10 +42,18 @@ export default class LedgerProvider extends Provider {
     let path = false
     let index = 0
 
-    while (!path) {
+    // A maximum number of addresses to lookup after which it is deemed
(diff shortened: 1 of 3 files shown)


src/providers/LedgerProvider.js
```diff
@@ -42,18 +43,31 @@ export default class LedgerProvider extends Provider {
...
+      throw new Error('Ledger: Wallet does not contain address')
+    }
+
     return path
   }
 
   async getAddressFromIndex (addressIndex) {
     const path = this.getDerivationPathFromIndex(addressIndex)
-    return this.getAddressFromDerivationPath(path)
+    if (path in this._addressCache) {
```
monokh, line 65: @harshjv  I eventually didn't cache it by first address. This is because it would then have to be implemented in `getAddresses`. Unfortunately though, we are using `getAddresses(0,1)` everywhere and always looking up the first address in this function would remove any performance benefit. 
We can hopefully mandate somehow that another device needs a new instance of the provider.

src/providers/bitcoin/BitcoinLedgerProvider.js
```diff
@@ -111,15 +110,17 @@ export default class BitcoinLedgerProvider extends LedgerProvider {
...
     let addressIndex = 0
     let currentAmount = 0
     let numOutputsOffset = 0
 
-    while ((currentAmount < amount) && maxAddresses > 0) {
-      const path = this.getDerivationPathFromIndex(addressIndex)
-      const address = await this.getAddressFromDerivationPath(path)
+    while ((currentAmount < amount)) {
+      const address = await this.getAddressFromIndex(addressIndex)
+      const isAddressUsed = await this.getMethod('isAddressUsed')(address.address)
```
harshjv: My only concern with this approach is that it requires now 2 network calls per loop, while earlier first 20 iteration required 1 network call.

Can we write a hybrid? Scan first 20 and then do this until we find an unused address?

| Addresses to be scanned | New approach | Hybrid approach |
|---|---|---|
| 5 addresses | 10 calls | 20 calls |
| 10 addresses | 20 calls | 20 calls |
| 20 addresses | 40 calls | 20 calls |
| 40 addresses | 80 calls | 60 calls |

harshjv (approved):
LGTM. Left a suggestion regarding scanning the addresses for utxos.

src/providers/bitcoin/BitcoinLedgerProvider.js
monokh: implemented that

PR #94, opened 2018-11-23
Bitcore provider

monokh:
@h4v0kr's BitcoreRPCProvider with some cleanup based #93


src/providers/bitcoin/BitcoinRPCProvider.js
```diff
@@ -1,7 +1,33 @@
...
+import BigNumber from 'bignumber.js'
 import JsonRpcProvider from '../JsonRpcProvider'
 
 export default class BitcoinRPCProvider extends JsonRpcProvider {
+  async signMessage (message, address) {
+    return new Promise((resolve, reject) => {
+      this.jsonrpc('signmessage', address, message).then(result => {
+        resolve(Buffer.from(result, 'base64'))
+      })
+    })
```
harshjv, line 11: or this can be just,

```
return this.jsonrpc('signmessage', address, message).then(result => Buffer.from(result, 'base64'))
```

harshjv (approved):
LGTM. A small improvement can be done.

PR #95, opened 2018-11-30
Normalise the bitcoin transaction calling and object

monokh:

- [x] Remove chain specific functions from Client class
- [x] Simplify and standardise retrieving transaction on bitcoin

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -274,38 +274,13 @@ export default class Client {
 
     const transaction = await this.getMethod('getTransactionByHash')(txHash)
 
-    const valid = this.validateTransaction(transaction)
+    if (transaction) {
+      const valid = this.validateTransaction(transaction)
 
-    if (!valid) {
-      const errors = this.validateTransaction.errors
-      throw new InvalidProviderResponseError(`Provider returned an invalid transaction: ${errors[0].dataPath} ${errors[0].message}`)
-    }
-
-    return transaction
-  }
-
-  /**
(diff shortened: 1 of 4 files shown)


src/Client.js
```diff
@@ -274,38 +274,13 @@ export default class Client {
...
 
-    if (!valid) {
-      const errors = this.validateTransaction.errors
-      throw new InvalidProviderResponseError(`Provider returned an invalid transaction: ${errors[0].dataPath} ${errors[0].message}`)
-    }
-
-    return transaction
-  }
-
-  /**
```
monokh, line 287: It didn't make much sense to have this in the client as it doesn't have an Ethereum implementation and there's little you can do with a raw transaction if not to decode it. also `getTransactionByHash` gives you back the tx hex anyway

matthewjablack: Not sure I agree with removing it just because it doesn't have an Ethereum implementation. Bitcoin developers may have a specific use case for it. Although since `getTransactionByHash` already has it, I think it makes sense 👍

matthewjablack (approved):
LGTM 🎉

src/Client.js
monokh: The client should only have cross chain methods otherwise it can become really cluttered. 
Solely bitcoin devs wouldn't find much use in CAL I think and if they wanted to use a bitcoin only implemented function, they can always use the provider directly.

PR #96, opened 2018-12-02
Validate network id MetaMask provider

matthewjablack:
### Description

Currently it's possible to have an interface where the rpc provider has one network id, and the wallet provider has a different network id, causing confusion. This PR ensures that there is a function to check if the network id is different between these providers. 

When users setup MetaMask provider, they can specify `networkId` if they want to, and validation will be done on `networkId` to ensure MetaMask is on the right network. 

`EthereumMetaMaskProvider(web3.currentProvider, 4)`

- [x] Add getWalletNetworkId

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -562,4 +562,19 @@ export default class Client {
 
     return this.getMethod('refundSwap')(initiationTxHash, recipientAddress, refundAddress, secretHash, expiration)
   }
+
+  async getWalletNetworkId () {
+    return this.getMethod('getWalletNetworkId')()
+  }
+
+  async getRPCNetworkId () {
+    return this.getMethod('getRPCNetworkId')()
+  }
+
+  async ensureNetworkId () {
+    const walletNetworkId = await this.getMethod('getWalletNetworkId')()
+    const rpcNetworkId = await this.getMethod('getRPCNetworkId')()
+
+    return walletNetworkId === rpcNetworkId
+  }
 }
--- a/src/Client.js
+++ b/src/Client.js
@@ -566,15 +566,4 @@ export default class Client {
   async getWalletNetworkId () {
     return this.getMethod('getWalletNetworkId')()
   }
-
-  async getRPCNetworkId () {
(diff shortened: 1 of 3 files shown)


monokh:
@mattBlackDesign the description makes it seem like this scenario is the fault of the library but it actually seems to be rather a misconfiguration by the user. The providers should use whichever network they are provided and making sure they match should be the responsibility (and also control) of the user. 

Maybe your issue is more that the metamask and rpc provider dont ensure that they are connecting to the network supplied in the constructor. In which case I think the better solution is to introduce the net version check in the constructor itself rather than a separate process.

matthewjablack:
@monokh agreed that this is not a fault of the library, but the issue I see with the argument of putting it in the constructor, is that it's easy for the user to change the network after the page has been loaded, which would end up with them using the wrong network in association with the data the node is sending them.

monokh:
@mattBlackDesign 
If we scope the issue down, it's to do with Metamask correct? i.e. you wouldn't know if the user changed the network setting on their metamask? For one thing, what i've noticed is that metamask force refreshes the page when you do that. Otherwise, it seems what we are saying is that a metamask provider initialised with a specific network ID shouldn't be able to be used to send things to metamask when it is switched to another network id. In that case, do metamask provider functions asserting the network ID before delegating solve the issue?

monokh:
Also note that in the proposed change, you would have to ensure your app is always handling this `ensureNetworkId` thing which depends on RPC endpoints being used at the same time (not a guarantee) whereas this should be something that is asserted by the provider itself too. Someone might be using only the wallet providers.

matthewjablack:
@monokh Metamask doesn't always refresh when changing network. 

Yep agreed on assumption that implementation is using both wallet and `rpc providers`. That's why this implementation gives them the option of getting `network id` from both, and then also a client implementation to check both. 

Perhaps there's a better way to ensure the calls are doing this. Maybe we can do this with `sendTransaction` and automatically detect if `rpc provider` is present

matthewjablack:
@monokh updated with validation at the `sendTransaction` level

PR #98, opened 2018-12-06
Get wallet info for balance and unused address

matthewjablack:
### Description

Add `getWalletInfo` for `LedgerProvider` and `MetaMask` to return `balance` and `unusedAddress`

- [x] `getWalletInfo` for `LedgerProvider`
- [x] `getWalletInfo` for `MetaMaskProvider`

### DIFF
--- a/src/providers/bitcoin/BitcoinLedgerProvider.js
+++ b/src/providers/bitcoin/BitcoinLedgerProvider.js
@@ -161,6 +161,35 @@ export default class BitcoinLedgerProvider extends LedgerProvider {
     }))
   }
 
+  async getWalletInfo (from = {}) {
+    let addressIndex = from.index || 0
+    let unusedAddress = false
+    let balance = 0
+
+    while (!unusedAddress) {
+      let addresses = await this.getAddresses(addressIndex, 10)
+      const addressList = addresses.map(addr => addr.address)
+
+      const addressDeltas = await this.getMethod('getAddressDeltas')(addressList)
+
+      addressDeltas.forEach((delta) => {
+        const addressIndex = addresses.findIndex(address => address.address === delta.address)
+        if (addresses[addressIndex].balance === undefined) { addresses[addressIndex].balance = 0 }
+        addresses[addressIndex].balance += delta.satoshis
+        balance += delta.satoshis
+      })
+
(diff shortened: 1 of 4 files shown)


src/providers/bitcoin/BitcoinLedgerProvider.js
```diff
@@ -161,6 +161,40 @@ export default class BitcoinLedgerProvider extends LedgerProvider {
...
+    let addressIndex = from.index || 0
+    let unusedAddress = false
+    let usedAddresses = []
+    let balance = 0
+
+    while (!unusedAddress) {
+      let addresses = await this.getAddresses(addressIndex, numAddressPerCall)
+      const addressList = addresses.map(addr => addr.address)
+
+      const addressDeltas = await this.getMethod('getAddressDeltas')(addressList)
```
monokh, line 174: @mattBlackDesign note that `getAddressDeltas` is a bitcore only function. With what we have before this, it will work both if you have a bitcoind or bitcore node because getUnspentTransactions is implemented in `BitcoinRPCProvider` and `BitcoreRPCProvider` but i don't see the same for `getAddressDeltas`.

src/providers/bitcoin/BitcoinLedgerProvider.js
```diff
@@ -161,6 +161,40 @@ export default class BitcoinLedgerProvider extends LedgerProvider {
...
+      addressDeltas.forEach((delta) => {
+        const addressIndex = addresses.findIndex(address => address.address === delta.address)
+        if (addresses[addressIndex].balance === undefined) { addresses[addressIndex].balance = 0 }
+        addresses[addressIndex].balance += delta.satoshis
+        balance += delta.satoshis
+      })
+
+      addresses.forEach((address) => {
+        if (!unusedAddress) {
+          if (address.balance === undefined) {
```
monokh, line 185: @mattBlackDesign is this right? Are we saying that an unused address is an address that does not have a balance? This doesn't seem right because you could have any number of addresses with 0 balance at the start of your ledger, (i know i do) before you hit one with a balance and as I understand here, unused address would come out wrong and potentially not calculate correct balance too.

In the other method we have used unspents or bitcore's `received` property.

PR #99, opened 2018-12-06
Use estimatesmartfee for bitcoin tx fee estimation

harshjv:


src/providers/bitcoin/BitcoinLedgerProvider.js
```diff
@@ -110,14 +110,17 @@ export default class BitcoinLedgerProvider extends LedgerProvider {
...
+  async getUtxosForAmount (amount) {
     const utxosToUse = []
     let addressIndex = 0
     let currentAmount = 0
     let numOutputsOffset = 0
 
     while ((currentAmount < amount)) {
-      const address = await this.getAddressFromIndex(addressIndex)
+      const [ feePerByte, address ] = await Promise.all([
+        this.getMethod('getFeePerByte')(2),
```
matthewjablack: Why 2 blocks and not 1?

src/providers/bitcoin/BitcoinRPCProvider.js
```diff
@@ -3,6 +3,10 @@ import BigNumber from 'bignumber.js'
 import JsonRpcProvider from '../JsonRpcProvider'
 
 export default class BitcoinRPCProvider extends JsonRpcProvider {
+  async getFeePerByte (numberOfBlocks = 2) {
+    return this.jsonrpc('estimatesmartfee', numberOfBlocks).then(({ feerate }) => (feerate * 1e8) / 1024)
```
matthewjablack, line 7: what's the calculation here?

matthewjablack (approved):
LGTM 🎉 just a couple comments.

src/providers/bitcoin/BitcoinRPCProvider.js
harshjv: `feerate` is in `BTC/kB`. While we are using `sat/b` in our calculations. So this is just converting `BTC -> sat` and then `kB -> b`

src/providers/bitcoin/BitcoinLedgerProvider.js
harshjv: It was just a default value i went with. Best was would be accepting user's preferred feerate/number of block from `sendTransaction` and `getUtxosForAmount`. We can look into that.

monokh (approved):
LGTM 🎉

src/providers/bitcoin/BitcoinRPCProvider.js
monokh: @harshjv  `feerate * 1e8` may create some floating point precision problems, may not matter that much but if you want a whole number, should probably use `BigNumber.js` for this calculation

PR #100, opened 2018-12-06
Improve finding swap transactions

monokh:
### Description

The method for finding the swap transactions would parse blocks and every transaction in that to find it. This was innefficient and on mainnet could make thousands of network calls (per tx) and take minutes

We need a more efficient way

- [x] Add `getAddressTransactions` to BitcoreRPCProvider - One call to get matching transactions in a block

### DIFF
--- a/examples/swap/swap-e2e.html
+++ b/examples/swap/swap-e2e.html
@@ -105,12 +105,12 @@
     /* global $, ChainAbstractionLayer, web3 */
     const { Client, providers, networks, crypto } = ChainAbstractionLayer
     const bitcoin = new Client()
-    bitcoin.addProvider(new providers.bitcoin.BitcoreRPCProvider('https://localhost:18332', 'bitcoin', 'local321'))
+    bitcoin.addProvider(new providers.bitcoin.BitcoreRPCProvider('http://localhost:18332', 'bitcoin', 'local321'))
     bitcoin.addProvider(new providers.bitcoin.BitcoinLedgerProvider({ network: networks.bitcoin_testnet, segwit: false }))
     bitcoin.addProvider(new providers.bitcoin.BitcoinSwapProvider({ network: networks.bitcoin_testnet }))
 
     const ethereum = new Client()
-    ethereum.addProvider(new providers.ethereum.EthereumRPCProvider('https://localhost:8545'))
+    ethereum.addProvider(new providers.ethereum.EthereumRPCProvider('http://localhost:8545'))
     ethereum.addProvider(new providers.ethereum.EthereumMetaMaskProvider(web3.currentProvider))
     ethereum.addProvider(new providers.ethereum.EthereumSwapProvider())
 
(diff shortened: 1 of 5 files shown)


src/providers/bitcoin/BitcoinRPCProvider.js
```diff
@@ -109,6 +109,20 @@ export default class BitcoinRPCProvider extends JsonRpcProvider {
     return this.jsonrpc('getblockcount')
   }
 
+  async getAddressTransactions (address, start, end) {
+    const transactionIds = []
+    for (const blockNumber of _.range(start, end + 1)) {
+      const block = await this.getMethod('getBlockByNumber')(blockNumber, true)
```
matthewjablack, line 115: Can't we use `getAddressUtxos` for this? Wouldn't this be easier than checking each block?

examples/swap/swap-e2e.html
```diff
@@ -105,12 +105,12 @@ <h3>Refund swap:</h3>
     /* global $, ChainAbstractionLayer, web3 */
     const { Client, providers, networks, crypto } = ChainAbstractionLayer
     const bitcoin = new Client()
-    bitcoin.addProvider(new providers.bitcoin.BitcoreRPCProvider('https://localhost:18332', 'bitcoin', 'local321'))
+    bitcoin.addProvider(new providers.bitcoin.BitcoreRPCProvider('http://localhost:18332', 'bitcoin', 'local321'))
```
matthewjablack, line 108: switch back to `https`

examples/swap/swap-e2e.html
```diff
@@ -105,12 +105,12 @@ <h3>Refund swap:</h3>
...
     bitcoin.addProvider(new providers.bitcoin.BitcoinLedgerProvider({ network: networks.bitcoin_testnet, segwit: false }))
     bitcoin.addProvider(new providers.bitcoin.BitcoinSwapProvider({ network: networks.bitcoin_testnet }))
 
     const ethereum = new Client()
-    ethereum.addProvider(new providers.ethereum.EthereumRPCProvider('https://localhost:8545'))
+    ethereum.addProvider(new providers.ethereum.EthereumRPCProvider('http://localhost:8545'))
```
matthewjablack, line 113: switch back to `https`

matthewjablack (approved):
LGTM 🎉 just some small comments

examples/swap/swap-e2e.html
monokh: I actually recall discussing whether https was required and i was under the impression that it was for some reason. I then tried to change to http and it still worked. So due to the default host port of starting a local rpc is http, i thought i could change it to that.

src/providers/bitcoin/BitcoinRPCProvider.js
monokh: `getAddressUtxos` is a bitcore function. This provider assumes bitcoind. 
we could do with `listunspent` but you need to make sure you are watching the p2sh so i don't see much point in that.

PR #103, opened 2018-12-09
Fix getUtxosForAmount to do batch address calls

matthewjablack:
### Description

This PR fixes `getUtxosForAmount` to do batch calls for addresses, and ensures that multiple utxos per address does not end up with unnecessary inputs. 

- [x] Modify `getUtxosForAmount` to do batch calls for addresses rather than check if each address is used


harshjv:
Closing because @h4v0kr has pulled these changes in https://github.com/liquality/chainabstractionlayer/tree/bitcore-provider manually.

PR #104, opened 2018-12-09
Fix findSwapTransaction to use address deltas

matthewjablack:
### Description

Find `findSwapTransaction` to use `getAddressDeltas` rather than relying on block

- [x] Modify `findSwapTransaction`


harshjv:
Closing because @h4v0kr has pulled these changes in https://github.com/liquality/chainabstractionlayer/tree/bitcore-provider manually.

PR #108, opened 2018-12-13
Add mempool txs for getUtxosForAmount and getUnusedAddress

matthewjablack:
### Description

This PR adds mempool txs for getUtxosForAmount and getUnusedAddress which allows for the creation of multiple swap transactions without `txn-mempool-conflict`

- [x] Add mempool txs for getUtxosForAmount and getUnusedAddress

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -400,8 +400,8 @@ export default class Client {
    * @param {!string} expiration - Expiration time
    * @return {Promise<string>} Resolves with a transaction identifier.
    */
-  async findInitiateSwapTransaction (value, recipientAddress, refundAddress, secretHash, expiration) {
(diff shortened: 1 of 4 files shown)


gr0kchain (commented):
Looks good, was successfully able to reproduce the error using master by calling doSwap() after initiateSwap in the cli/bitcoin.js.

RpcError: 258: txn-mempool-conflict

After updating to this PR the error above does not occur and swaps are executed as expected. A note, that running multiple version concurrently using this method yields a device locking problem on Ledger. The current code base does not handle exceptions with device locks and assumes getApp in the LedgerProvider always returns a transport.

Please ensure these PR's are made to development and not master.

gr0kchain:
I've created a development branch from Master, and merged this pr with it. Closing

PR #112, opened 2018-12-13
Fix swap mempool txs

monokh:
- [x] Ensure mempool transactions are added to list before full tx objects are retrieved
- [x] Remove unused `includemempool` param


matthewjablack (approved):
LGTM 🎉

PR #117, opened 2018-12-17
Use numAddressPerCall for getUnusedAddress

harshjv:


harshjv:
Closing because of #119

PR #121, opened 2018-12-18
Cache xpubkey

monokh:
### Description

Cache the xpubkey by path. This prevents too many unneeded calls to export pub key to the ledger. 
For a typical use cases reduces 8 calls to export key to 2 
2 is required because the paths differ and have to be individually exported. 

- [x] Cache the xpubkey when it is generated
- [x] Fix the browser dev webpack config so it includes buffer


harshjv (approved):
Nice!

PR #123, opened 2018-12-18
implement address gap for getUnusedAddress

harshjv:


src/providers/bitcoin/BitcoinLedgerProvider.js
```diff
@@ -450,22 +450,28 @@ export default class BitcoinLedgerProvider extends LedgerProvider {
...
-  async getUsedAddresses (numAddressPerCall = 100) {
+  // addressType
+  // nonChange: 0
+  // change: 1
+  // both: 2
+  async _getUsedUnusedAddresses (numAddressPerCall = 100, addressType) {
     const addressGap = 20
-    const usedAddress = []
+    const usedAddresses = []
+    const uacMap = { change: 0, nonChange: 0 }
```
matthewjablack: What does uacMap stand for?

matthewjablack (approved):
One small comment. Other than that LGTM 🎉

src/providers/bitcoin/BitcoinLedgerProvider.js
harshjv: Earlier it stood for `Unused Address Change Map` and then lately i've been using it for a different purpose. I'll update it.

PR #125, opened 2018-12-19
Optimize and fix getDerivationPathFromAddress

monokh:

- [x] Fixes getDerivationPathFromAddress so that it does not stop at 50 addresses - instead 1000
- [x] Optimises getDerivationPathFromAddress to use batching address gets and return early

### DIFF
--- a/src/providers/LedgerProvider.js
+++ b/src/providers/LedgerProvider.js
@@ -42,29 +42,28 @@ export default class LedgerProvider extends Provider {
   }
 
   async getDerivationPathFromAddress (address) {
-    let path = false
     let index = 0
     let change = false
 
     // A maximum number of addresses to lookup after which it is deemed
     // that the wallet does not contain this address
-    const maxAddresses = 50
+    const maxAddresses = 1000
+    const addressesPerCall = 50
 
-    while (!path && index < maxAddresses) {
-      const addr = await this.getAddresses(index, 1)
-      if (addr[0].address === address) path = this.getDerivationPathFromIndex(index, change)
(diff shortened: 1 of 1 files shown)


src/providers/LedgerProvider.js
```diff
@@ -42,29 +42,28 @@ export default class LedgerProvider extends Provider {
...
 
     // A maximum number of addresses to lookup after which it is deemed
     // that the wallet does not contain this address
-    const maxAddresses = 50
-
-    while (!path && index < maxAddresses) {
-      const addr = await this.getAddresses(index, 1)
-      if (addr[0].address === address) path = this.getDerivationPathFromIndex(index, change)
-      index++
+    const maxAddresses = 1000
```
matthewjablack, line 50: Hey @monokh do you think we need a max number of addresses to look through? Are we assuming a max number of transactions before the ledger is unusable.

Perhaps we should pair this with balance of addresses, and only get addresses that have a balance or are the first unused address.

Looks good for now though  🙂

matthewjablack (approved):

src/providers/LedgerProvider.js
monokh: We need a max number to be able to break this function if we cannot find it in the ledger. 

We could use balance but there would be a huge number of network calls involved if this function was run in isolation. Also, pairing it with the balance functions in a way to reuse the network calls would make this pretty spaghetti i would think. Address generation is almost 0 cost so I wouldn't worry about even increase the max here to something very high like 100,000  if we need to.

PR #131, opened 2019-01-04
Separate networks

monokh:
- [x] Separate the network configs for ethereum and bitcoin
- [x] Add an `isTestnet` flag to each network descriptor
- [x] Change EthereumMetaMaskProvider to accept a network config rather than just a plain network ID


examples/swap/cli/getsecret.js
```diff
@@ -1,5 +1,6 @@
 var ChainAbstractionLayer = require('../../../dist/index.cjs.js')
-const { Client, providers, networks, crypto } = ChainAbstractionLayer
+const { Client, providers, crypto } = ChainAbstractionLayer
+const networks = providers.bitcoin.networks
```
monokh, line 3: The structure is a bit ugly atm. This would be nicer: 
`bitcoin.providers`
`bitcoin.networks`
a big restructure would be needed though, which is why i've not done this in this PR.

harshjv (approved):
LGTM

PR #132, opened 2019-01-11
E2E Swap tests

monokh:
### Description

Adding e2e test of swap functionality. 

- [x] BitcoreRPCProvider: Remove duplicate methods already in BitcoinRPCProvider
- [x] RPC providers: fix format of data returned from getUnusedAddress
- [x] BitcoinJsLibSwapProvider: fix redeeming and setting network
- [x] Tests for single chain flow - node and wallets
- [x] Tests for BTC-ETH swap using node and wallets
- [x] Readme explaining how to run tests


package.json
```diff
@@ -34,7 +34,8 @@
   "scripts": {
     "cover": "nyc --reporter=html --reporter=text npm test",
     "lint": "standard \"**/*.js\" \"**/*.html\" \"**/*.md\"",
-    "test": "cross-env NODE_ENV=test mocha $npm_package_options_mocha",
+    "test": "cross-env NODE_ENV=test mocha test/unit $npm_package_options_mocha",
```
harshjv, line 37: i am directly using `nyc` so with test, we can also see coverage.

https://github.com/liquality/chainabstractionlayer/commit/103fbe94d6cdbf0d61533b04ad8f25ef370aa77c#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R37

test/integration/swap/common.js
```diff
@@ -0,0 +1,81 @@
+import { expect } from 'chai'
+import MetaMaskConnector from 'node-metamask'
+import { Client, providers, crypto } from '../../../'
```
harshjv: Use files from `src` (instead of `dist`) so it can be reported back to coveralls.

harshjv (changes requested):
Left some comments. Haven't tried yet.

package.json
monokh: The purpose of these tests are only to run locally so it's kind of strange to reflect the code coverage in it. It wouldn't be relevant to someone running it.
Perhaps we need an overall `test` script that runs `nyc, npm run test:unit, npm run test:swap` ?

test/integration/swap/common.js
monokh: changed.

harshjv (approved):
LGTM

PR #134, opened 2019-01-25
Fix BitcoinLedgerProvider retrieving public key from address not using cache

monokh:
### Description

The swap provider used the `getPubKey` method to retrieve the public key of a ledger address, this method needed to retrieve data from the ledger directly which caused unnecessary calls. This data can be derived from the public key previously exported from the ledger.

- [x] Refactor swap provider to use the addresses that can be generated from the pub key
- [x] Use `bip32.fromPublicKey` to derive paths directly from the public key and chain code. This removes the need to generate an xpubkey ourselves.
- [x] `getDerivationPathFromAddress` is now `getWalletAddress` which can be used to retrieve all information (including derivation path and public key) about an address from the ledger.
- [x] Removed some dead code

### DIFF
--- a/src/providers/bitcoin/BitcoinJsLibSwapProvider.js
+++ b/src/providers/bitcoin/BitcoinJsLibSwapProvider.js
@@ -1,5 +1,5 @@
 import Provider from '../../Provider'
-import { addressToPubKeyHash, compressPubKey, pubKeyToAddress, reverseBuffer, scriptNumEncode } from './BitcoinUtil'
+import { addressToPubKeyHash, pubKeyToAddress, reverseBuffer, scriptNumEncode } from './BitcoinUtil'
 import { sha256, padHexStart } from '../../crypto'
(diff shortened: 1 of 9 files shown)


src/providers/bitcoin/BitcoinJsLibSwapProvider.js
```diff
@@ -102,42 +102,7 @@ export default class BitcoinJsLibSwapProvider extends Provider {
   }
 
   async refundSwap (initiationTxHash, recipientAddress, refundAddress, secretHash, expiration) {
-    return this._redeemSwap(initiationTxHash, recipientAddress, refundAddress, secretHash, expiration, false)
-  }
-
-  async _redeemSwap (initiationTxHash, recipientAddress, refundAddress, secretParam, expiration, isClaim) {
```
monokh, line 108: there were some references to old code here. Since the code is just copied, thought i'd remove it

PR #135, opened 2019-01-30
Wallet providers assert matching network

monokh:
### Description

Wallet providers need to have a method of determining whether they are on the right network. Also, this should be asserted before sending a request to the wallet. 

- [x] Set in place the interface for wallet providers.
- [x] Add mechanism for Wallet providers to always check the network before calls. 
- [x] Implemented `getConnectedNetwork` for metamask provider
- [x] Implementation for `LedgerProvider` ?


gr0kchain (approved):
This looks good

PR #139, opened 2019-02-12
BitcoinSwapProvider: Not enough value claim error

monokh:
### Description

When the value locked in the swap is smaller than the fee required to send the transaction, the library does not fail and continues to request a signature from the ledger for a negative amount. 
This PR ensures that an error is thrown in that case.

### DIFF
--- a/src/providers/bitcoin/BitcoinSwapProvider.js
+++ b/src/providers/bitcoin/BitcoinSwapProvider.js
@@ -228,8 +228,14 @@ export default class BitcoinSwapProvider extends Provider {
     const output = initiationTx.outputs[voutIndex]
     const value = parseInt(reverseBuffer(output.amount).toString('hex'), 16)
     const { relayfee } = await this.getMethod('jsonrpc')('getinfo')
-    const fee = this.getMethod('calculateFee')(1, 1, feePerByte)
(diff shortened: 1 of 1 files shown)


src/providers/bitcoin/BitcoinSwapProvider.js
```diff
@@ -228,8 +228,14 @@ export default class BitcoinSwapProvider extends Provider {
     const output = initiationTx.outputs[voutIndex]
     const value = parseInt(reverseBuffer(output.amount).toString('hex'), 16)
     const { relayfee } = await this.getMethod('jsonrpc')('getinfo')
-    const fee = this.getMethod('calculateFee')(1, 1, feePerByte)
-    const amount = value - Math.max(fee, BigNumber(relayfee).times(1e8).toNumber())
+    const calculatedFee = this.getMethod('calculateFee')(1, 1, feePerByte)
+    const fee = Math.max(calculatedFee, BigNumber(relayfee).times(1e8).toNumber())
```
gr0kchain: Since we're doing BigNumber calcs, should we not use BigNumber.max, and then also keep doing the math as value.minus(fee) 

Somehow feel uneasy about mixing BigNumbers with built-in data types

monokh: Good point! fixed

PR #141, opened 2019-02-14
Remove unused code

monokh:
### Description

Removing unused code

### Note worthy deletions:

- [x] string to provider converter (DSNProvider) - unused. We should only maintain one way of adding providers
- [x] `getDerivationPathFromIndex` and `getAddressFromIndex`. Unused. these use cases are covered by `getAddresses`
- [x] unused code from `BitcoinJsLibSwapProvider`


src/Client.js
```diff
@@ -43,36 +40,12 @@ export default class Client {
...
   /**
    * Add a provider
-   * @param {!string|Provider} provider - The provider instance or RPC connection string
+   * @param {!Provider} provider - The provider instance or RPC connection string
    * @return {Client} Returns instance of Client
    * @throws {InvalidProviderError} When invalid provider is provider
    * @throws {DuplicateProviderError} When same provider is added again
    */
   addProvider (provider) {
-    if (isString(provider)) {
```
gr0kchain, line 52: I'd like to keep this [proto]://[user]:[pass]@[host]:[port]

gr0kchain (commented):
Looks good, just added a note on the string DSN feature

src/Client.js
monokh: Is there a particular reason? 
As well as having to test and support this functionality going forward, it's very specialised:
- It only works for RPC providers
- It only works for RPC providers provided internally by the library that follow the same argument order
- It does not support providing any additional arguments that an RPC provider might take

gr0kchain: I guess we can remove it if there is no use

PR #148, opened 2019-02-27
Added fix for unit test

gr0kchain:
Added fixes for BIP199 + checking the size of the secret is always 32 bytes, Also updated the secret extraction from the redeem tx and the unit test for createSwapScript. Note that an update to bitcoinjs-lib has also been made and now uses "bitcoinjs-lib": "https://github.com/h4v0kr/bitcoinjs-lib"


monokh (approved):
LGTM, please merge to dev so we can run the test suite

PR #151, opened 2019-03-13
Use coinselect for finding utxos

harshjv:
- [x] Use https://github.com/bitcoinjs/coinselect for calculating change & utxos
- [x] Move `calculateFee` to BitcoinUtil

### DIFF
--- a/src/providers/bitcoin/BitcoinLedgerProvider.js
+++ b/src/providers/bitcoin/BitcoinLedgerProvider.js
@@ -1,11 +1,15 @@
 import LedgerProvider from '../LedgerProvider'
 import Bitcoin from '@ledgerhq/hw-app-btc'
 
+import { find } from 'lodash'
 import { BigNumber } from 'bignumber.js'
 import { base58, padHexStart } from '../../crypto'
 import { pubKeyToAddress, addressToPubKeyHash, compressPubKey, getAddressNetwork } from './BitcoinUtil'
 import networks from './networks'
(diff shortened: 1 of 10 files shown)


src/providers/bitcoin/BitcoinLedgerProvider.js
```diff
@@ -142,7 +131,49 @@ export default class BitcoinLedgerProvider extends LedgerProvider {
...
+          const addr = find(addrList, addr => addr.address === utxo.address)
+
+          utxo.value = utxo.satoshis
+          utxo.derivationPath = addr.derivationPath
+
+          return utxo
+        })
+
+      globalUtxoSet = globalUtxoSet.concat(utxos)
+      if (feePerByte === false) feePerByte = await feePerBytePromise
```
monokh, line 159: confused why the promise isn't just called here instead of using a variable?

monokh (approved):
LGTM! minor comment
Just checking you have manage to run the integration tests to ensure the ledger swaps continue to work?

src/providers/bitcoin/BitcoinLedgerProvider.js
harshjv: We are initiating `getFeePerByte` on line 107 and after that we are also making other network calls. While `getFeePerByte` isn't required until line 161, it is essentially being called in parallel in a non-blocking way.

harshjv:
@monokh all green!

PR #155, opened 2019-03-20
Use webusb transport for Ledger

harshjv:
- [x] Switch to webusb
- [x] Update all the Ledgerjs libs to 4.46.x
- [x] Update all babel libs to 7.4.x


src/providers/LedgerWebTransport.js
```diff
@@ -0,0 +1,7 @@
+import WebUSBTransport from '@ledgerhq/hw-transport-webusb'
```
monokh, line 1: What's the purpose of this extra class?

monokh (approved):
LGTM
not sure why we need the extra class

src/providers/LedgerWebTransport.js
harshjv: It adds extra static member `isWebUSBTransport` to the class, so we can check & run claim/release logic just for the browser (webusb) and not in node (hid).

PR #162, opened 2019-03-28
Swap: add startBlock parameter to find functions

monokh:
The `startBlock` parameter for the find functions will allow us to within reasonable time find any swap transaction when we know which block it was initiated at. 
This is important for when we don't have the transactions or looking for arbitary swaps.


harshjv (approved):
LGTM

PR #163, opened 2019-03-28
EthereumMetaMaskProvider: Ensure access is granted before requesting

monokh:
https://medium.com/metamask/https-medium-com-metamask-breaking-change-injecting-web3-7722797916a8

Enable access to metamask before requesting. When privacy mode is enabled, this will be absolutely required.


harshjv (commented):
LGTM

PR #165, opened 2019-04-02
Ethereum erc20

grandsmarquis:
### Description

Add ERC20 Swap provider.

- [x] Enables Transfer, balance of ERC20 tokens
- [x] Enables Swaps of ERC20 tokens


grandsmarquis:
for runing test on Ganache there is a bug and need to add 
```
      gasPrice: 8000000000,
      gas: 4700000
```

in the EthereumRPCProvider line 38

src/providers/ethereum/EthereumERC20Provider.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  generateErc20Transfer (to, value) {
+    const functionSignature = 'a9059cbb'
+    const encodedAddress = padHexStart(ensureAddressStandardFormat(to), 64)
+    const encodedValue = padHexStart(value.toString(16), 64)
+    return functionSignature + encodedAddress + encodedValue
+  }
+
+  erc20Transfer (to, value) {
```
monokh, line 18: Isn't this essentially an implementation of `sendTransaction`? if you have added an ERC20 provider to your client, you want to be able to do `clientErc20.sendTransaction('0x0000', 10000000)` and it will transfer the erc20. Using the standard interface. 
By the way, you can still call `sendTransaction` within the function, it will look through the stack of providers and find the next in line.

src/providers/ethereum/EthereumERC20Provider.js
```diff
@@ -0,0 +1,76 @@
...
+    const txBytecode = this.generateErc20Transfer(to, value)
+    return this.getMethod('sendTransaction')(ensureHexEthFormat(this._contractAddress), 0, txBytecode)
+  }
+
+  // TODO temporary way to expose contract address to other providers
+  erc20 () {
```
monokh: Are these 2 functions necessary? Seems like we can just access this._contractAddress in any case.

src/providers/ethereum/EthereumERC20Provider.js
```diff
@@ -0,0 +1,76 @@
...
+    const functionSignature = '0x70a08231'
+    addresses = addresses
+      .map(address => String(address))
+    const addrs = addresses.map(ensureHexEthFormat)
+    const promiseBalances = await Promise.all(addrs.map(address => this.getMethod('jsonrpc')('eth_call', { data: functionSignature + padHexStart(ensureAddressStandardFormat(address), 64), to: ensureHexEthFormat(this._contractAddress) }, 'latest')))
+    return promiseBalances.map(balance => parseInt(balance, 16))
+      .reduce((acc, balance) => acc + balance, 0)
+  }
+
+  async getBalance (addresses) {
```
monokh: Bit of an odd construct. If it just relays the call to erc20Balance, why not just include the content of that function here?

src/providers/ethereum/EthereumERC20Provider.js
```diff
@@ -0,0 +1,76 @@
...
+    return this.erc20Balance(addresses)
+  }
+
+  // TODO not used anymore (was only there for tests)
```
monokh: Please remove the functions if not used 😃 can always add back later if needed

src/providers/ethereum/EthereumERC20SwapProvider.js
```diff
@@ -0,0 +1,130 @@
+import Provider from '../../Provider'
+import { padHexStart } from '../../crypto'
+import { ensureAddressStandardFormat } from './EthereumUtil'
+
+const sleep = ms => new Promise(resolve => setTimeout(resolve, ms))
+
+// TODO Get contract address from constructor
+export default class EthereumERC20SwapProvider extends Provider {
+  createSwapScript (recipientAddress, refundAddress, secretHash, expiration) {
+    /* Source of the compiled Bytecode
```
monokh, line 8: Nice one, i think this is a good to be more transparent

src/providers/ethereum/EthereumERC20SwapProvider.js
```diff
@@ -0,0 +1,130 @@
...
+    return bytecode
+  }
+
+  async initiateSwap (value, recipientAddress, refundAddress, secretHash, expiration) {
+    const bytecode = this.createSwapScript(recipientAddress, refundAddress, secretHash, expiration)
+    await this.getMethod('sendTransaction')(null, 0, bytecode)
```
monokh: Referring to my previous comment about implementing erc20transfer as `sendTransaction`, i can see why this might not work now...

src/providers/ethereum/EthereumERC20SwapProvider.js
```diff
@@ -0,0 +1,130 @@
...
+    const tx = await this.findContractDeployTransaction(value, recipientAddress, refundAddress, secretHash, expiration)
+    const initiationTransactionReceipt = await this.getMethod('getTransactionReceipt')(tx.hash)
+    await this.getMethod('erc20Transfer')(initiationTransactionReceipt.contractAddress, value)
+    let contractBalanceMatchesValue = false
```
monokh: Why does initiating need to wait for the balance? When waiting for initiation to complete, someone would just use `findInitiateSwapTransaction` i would think. That's how its' generally done with the other assets

src/providers/ethereum/EthereumERC20SwapProvider.js
```diff
@@ -0,0 +1,130 @@
...
+  async doesBalanceMatchValue (contractAddress, value) {
+    const balance = await this.getMethod('erc20Balance')([contractAddress])
+    return balance === value
+  }
+
+  async verifyInitiateSwapTransaction (initiationTxHash, value, recipientAddress, refundAddress, secretHash, expiration) {
+    const initiationTransaction = await this.getMethod('getTransactionByHash')(initiationTxHash)
+    const initiationTransactionReceipt = await this.getMethod('getTransactionReceipt')(initiationTxHash)
+    const transactionMatchesSwapParams = this.doesTransactionMatchSwapParams(initiationTransaction, recipientAddress, refundAddress, secretHash, expiration)
+    return transactionMatchesSwapParams && initiationTransactionReceipt.status === '1'
```
monokh: The balance should be checked here i would think. 
Doesn't need to scan for it though

test/integration/swap/singleChain.js
```diff
@@ -190,6 +190,17 @@ describe('Swap Single Chain Flow', function () {
     testSingle(chains.ethereumWithNode)
   })
 
+  describe('ERC20 - MetaMask', () => {
```
monokh, line 197: I know you don't have a ledger atm, but even if you can't test it, could you add the one with ledger too?

test/integration/swap/common.js
```diff
@@ -129,14 +144,36 @@ function mineBitcoinBlocks () {
...
+      if (block) {
+        tokenDeployedTransaction = block.transactions.find(transaction =>
+          transaction.input === bytecode
+        )
+        blockNumber++
+      }
+      await sleep(5000)
+    }
+    const initiationTransactionReceipt = await client.getMethod('getTransactionReceipt')(tokenDeployedTransaction.hash)
+    client.getMethod('setErc20')(initiationTransactionReceipt.contractAddress)
```
monokh: I would rather use some hacky way of setting the `._contractAddress` of the provider - like iterating through the providers and setting it, rather than have an exposed method on the client for that purpose

test/integration/swap/common.js
```diff
@@ -129,14 +144,36 @@ function mineBitcoinBlocks () {
   }
 }
 
-function connectMetaMask () {
+function connectMetaMask (client) {
```
monokh: Doesn't seem to be using the `client`

monokh (changes requested):
Great work. some suggestions

src/providers/ethereum/EthereumERC20Provider.js
```diff
@@ -0,0 +1,76 @@
+import Provider from '../../Provider'
+import { padHexStart } from '../../crypto'
+import { ensureAddressStandardFormat, ensureHexEthFormat } from './EthereumUtil'
+
+const sleep = ms => new Promise(resolve => setTimeout(resolve, ms))
```
harshjv: We are using this code very frequently in our codebase. Can we move it to a utility file?

src/providers/ethereum/EthereumRPCProvider.js
```diff
@@ -44,11 +44,6 @@ export default class EthereumRPCProvider extends JsonRpcProvider {
     return ensureHexStandardFormat(txHash)
   }
 
-  async sendRawTransaction (hash) {
```
monokh: nooooooo, we need these :( accident maybe?

monokh (changes requested):
everything looking good! i just think i spotted an accident with git

monokh (approved):
LGTM 🎉 🎈 🇪🇹 !!!!

PR #166, opened 2019-04-07
Add generate secrets fn

matthewjablack:
### Description

This PR adds the function `generateSecrets` to `Client.js`. This allows a user to sign a single message, and generate as many secrets as they want safely and deterministically, by simply incrementing the hash returned from signing and hashing that to get the secret. 

This is a much better user experience for cross-chain applications that require generating multiple secrets. 

- [x] Add `generateSecrets` fn to `Client.js`

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -1,5 +1,6 @@
 import { find, findLast, findLastIndex, isArray, isBoolean, isFunction, isNumber, isString } from 'lodash'
 import Ajv from 'ajv'
+import BigNumber from 'bignumber.js'
 
 import { Block, Transaction } from './schema'
 import { sha256 } from './crypto'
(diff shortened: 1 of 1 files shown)


monokh (commented):
@mattblackdesign  could you elaborate what the advantage of this would be over calling `generateSecret` multiple times? Even could the for loop be implemented by the user of the library instead? The issue I can see is we will have multiple ways of doing the same thing and I believe the secrets coming out of this have different security characteristics than calling `generateSecret` multiple times

matthewjablack:
Great points @monokh. I think this can be optimized to ensure that `signedMessage` can't be derived from `signedMessage` + 100. Closing for now until we have a new implementation.

matthewjablack:
@monokh changed the derivation process to ensure it is not possible to derive the message used to generate the secret from the 100th derived message.

monokh:
I think we agreed we could avoid this for now? Feel free to reopen otherwise

PR #172, opened 2019-04-12
Add edge case for getBalance where tokenOwner has no balance

matthewjablack:
### Description

This PR adds an edge case for `getBalance` where `tokenOwner` has no balance. Currently this returns `NaN` but it should really return 0

- [x] Add edge case for `getBalance` of `tokenOwner`

### DIFF
--- a/src/providers/ethereum/EthereumERC20Provider.js
+++ b/src/providers/ethereum/EthereumERC20Provider.js
@@ -32,8 +32,9 @@ export default class EthereumERC20Provider extends Provider {
     addresses = addresses
       .map(address => String(address))
     const addrs = addresses.map(ensureHexEthFormat)
-    const promiseBalances = await Promise.all(addrs.map(address => this.getMethod('jsonrpc')('eth_call', { data: functionSignature + padHexStart(ensureAddressStandardFormat(address), 64), to: ensureHexEthFormat(this._contractAddress) }, 'latest')))
-    return promiseBalances.map(balance => new BigNumber(balance, 16))
-      .reduce((acc, balance) => acc.plus(balance), new BigNumber(0))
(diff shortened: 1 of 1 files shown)


src/providers/ethereum/EthereumERC20Provider.js
```diff
@@ -32,8 +32,9 @@ export default class EthereumERC20Provider extends Provider {
     addresses = addresses
       .map(address => String(address))
     const addrs = addresses.map(ensureHexEthFormat)
-    const promiseBalances = await Promise.all(addrs.map(address => this.getMethod('jsonrpc')('eth_call', { data: functionSignature + padHexStart(ensureAddressStandardFormat(address), 64), to: ensureHexEthFormat(this._contractAddress) }, 'latest')))
-    return promiseBalances.map(balance => new BigNumber(balance, 16))
-      .reduce((acc, balance) => acc.plus(balance), new BigNumber(0))
+    const balances = (await Promise.all(addrs.map(address => this.getMethod('jsonrpc')('eth_call', { data: functionSignature + padHexStart(ensureAddressStandardFormat(address), 64), to: ensureHexEthFormat(this._contractAddress) }, 'latest'))))
+      .map(balance => parseInt(balance, 16))
```
harshjv, line 36: There's a catch here. If intermediate address has no balance but one of the addresses has some balance, this snippet will still return 0.

Instead, if we can check for NaN for each balance and then use `BigNumber` for each (for consistency)

Something like;

    .map(balance => BigNumber(balance, 16))
    .filter(balance => !balance.isNaN())
    .reduce((acc, balance) => acc.plus(balance), BigNumber(0))

harshjv (changes requested):

grandsmarquis (approved):
Mice catch!

harshjv:
See #182.

PR #175, opened 2019-04-23
Ethereum & Bitcoin node signing + Ledger 2 Ledger

grandsmarquis:
### Description

This enables node signing for both chains and ease ledger to ledger swaps.

### DIFF
--- a/src/providers/LedgerProvider.js
+++ b/src/providers/LedgerProvider.js
@@ -23,20 +23,21 @@ export default class LedgerProvider extends WalletProvider {
   }
 
   async createTransport () {
-    if (!LedgerProvider.transport) {
+    if (!this._transport) {
       debug('creating ledger transport')
-      LedgerProvider.transport = await Transport.create()
+      this._transport = await Transport.create()
       debug('ledger transport created')
-      LedgerProvider.transport.on('disconnect', () => {
+      this._transport.on('disconnect', () => {
         debug('ledger disconnected')
         this._appInstance = null
-        LedgerProvider.transport = null
+        this._transport = null
       })
     }
   }
 
   errorProxy (target, func) {
     const method = target[func]
+    const ctx = this
     if (Object.getOwnPropertyNames(target).includes(func) && typeof method === 'function') {
       return async (...args) => {
         debug(`calling "${func}" on ledger object`, args)
@@ -49,6 +50,7 @@ export default class LedgerProvider extends WalletProvider {
           return result
         } catch (e) {
           const { name, ...errorNoName } = e
+          ctx._transport = null
           throw new WalletError(e.toString(), errorNoName)
         }
       }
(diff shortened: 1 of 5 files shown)


src/providers/LedgerProvider.js
```diff
@@ -23,33 +23,35 @@ export default class LedgerProvider extends WalletProvider {
...
           debug(`result from "${func}" on ledger object`, result)
-
           return result
         } catch (e) {
-          const { name, ...errorNoName } = e
-          throw new WalletError(e.toString(), errorNoName)
+          // const { name, ...errorNoName } = e
+          ctx._transport = null
+          ctx._appInstance = null
+          // throw new WalletError(e.toString(), errorNoName)
```
harshjv: Is it okay to return null instead of throwing an error? This will simply return a null value if any ledger function fails, and it will make checking for null mandatory everywhere it is used, right?

src/providers/LedgerProvider.js
```diff
@@ -23,33 +23,35 @@ export default class LedgerProvider extends WalletProvider {
...
       debug('creating ledger transport')
-      LedgerProvider.transport = await Transport.create()
+      this._transport = await Transport.create()
       debug('ledger transport created')
-      LedgerProvider.transport.on('disconnect', () => {
+      this._transport.on('disconnect', () => {
         debug('ledger disconnected')
         this._appInstance = null
-        LedgerProvider.transport = null
+        this._transport = null
```
harshjv, line 33: Earlier we've used static property because the transport is independent of ledger apps. Can you confirm that this change is necessary?

harshjv (changes requested):

src/providers/LedgerProvider.js
monokh: Ageed. The purpose of this was so that `isWalletAvailable` returns false instead of an error? I think we discussed still throwing the error but having the first line in `isWalletAvailable` catch it and then return false.

src/providers/bitcoin/BitcoinRPCProvider.js
```diff
@@ -144,6 +144,22 @@ export default class BitcoinRPCProvider extends JsonRpcProvider {
     }
   }
 
+  async getUsedAddresses () {
+    const addresses = await this.jsonrpc('listaddressgroupings')
```
monokh, line 148: As we discussed, this would have to return every address ever generated by the node.

src/providers/LedgerProvider.js
grandsmarquis: In my tests I found out the once the transport is initiated with one app we don't manage to reuse the same with another app

src/providers/LedgerProvider.js
grandsmarquis: Agree and throwing again and catching inside of `isWalletAvailable`

harshjv (approved):
LGTM

monokh (approved):

PR #177, opened 2019-05-02
Fixes for Ethereum Ledger provider and Geth

grandsmarquis:
### Description

Correct some RPC calls to enable swap through geth clients and Ledger device


monokh (approved):
LGTM, 
I'm surprised how the ledger signed transaction can be invalid to some nodes.

PR #182, opened 2019-05-15
Fix sendTransaction in ERC20 provider

harshjv:
- [x] Fix `sendTransaction` in ERC20 provider


packages/ethereum-erc20-provider/lib/EthereumErc20Provider.js
```diff
@@ -4,8 +4,8 @@ import { BigNumber } from 'bignumber.js'
 import Provider from '@liquality/provider'
 import { padHexStart } from '@liquality/crypto'
 import {
-  ensureAddressStandardFormat,
-  ensureHexEthFormat
+  toLowerCaseWithout0x,
```
monokh: Yea this is kind of what I meant when i said the naming might start to mean more than just the 0x 😄 
I hope we don't start to have sentences describing what a function does eventually... This is why calling it just `standard` or `normalised` is a better naming in my opinion.

packages/ethereum-erc20-provider/lib/EthereumErc20Provider.js
```diff
@@ -4,8 +4,8 @@ import { BigNumber } from 'bignumber.js'
...
+  ensure0x
```
monokh: Note that eventually this might not even be just 0x, but also [EIP-55] (https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md) in which case we need to update these names again. Just another reason i think giving it a name rather than describing it is better

PR #184, opened 2019-05-19
Atomic Loans Additions to CAL with LoanFund and Loan Provider

matthewjablack:
### Description

Atomic Loans additions such as `LoanFund` Provider and `Loan` Provider

Introduce `getMempoolBalance`

- [ ] LoanFund Provider


packages/bitcoin-bitcoinjs-lib-swap-provider/package.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "@liquality/bitcoin-bitcoinjs-lib-swap-provider",
```
monokh: I think there are some unneeded changes here? 😄

packages/client/lib/Chain.js
```diff
@@ -180,6 +180,14 @@ export default class Chain {
     return balance
   }
 
+  async getMempoolBalance (addresses) {
```
monokh, line 183: These methods shouldn't be in the interface as they are not cross chain. Are you having issues accessing them internally in the providers using `getMethod`?

packages/client/lib/Client.js
```diff
@@ -148,6 +152,14 @@ export default class Client {
   get collateral () {
     return this._collateral
   }
+
+  get loan () {
```
monokh, line 156: what's the difference with the `collateral` side? It seems we are expanding the namespaces quite quickly for your usecase. Perhaps it could all come under `loan` and then you can have sub namespaces like `loan.fund.xxxxxx``loan.collateral.xxxxxx` ?

packages/ethereum-erc20-provider/lib/EthereumErc20Provider.js
```diff
@@ -76,6 +90,36 @@ export default class EthereumErc20Provider extends Provider {
...
+    addresses = addresses
+      .map(address => String(address))
+    const addrs = addresses.map(ensure0x)
+    const balances = (await Promise.all(addrs.map(address => this.getMethod('jsonrpc')('eth_call', { data: functionSignature + padHexStart(removeAddress0x(address), 64), to: ensure0x(this._contractAddress) }, 'pending'))))
+      .map(balance => parseInt(balance, 16))
+      .reduce((acc, balance) => acc + balance, 0)
+    return isNaN(balances) ? 0 : balances
+  }
+
+  async getErc20Name () {
```
monokh, line 106: Not really sure why we need this sort of functionality within the CAL. They cannot be replicated on any other chain therefore I think whichever the standard way of doing this in ethereum world would be a better approach than implementing all of this into the CAL.

monokh (changes requested):
Some comments.
Since you are now looking to add quite a bit of code to the CAL, I think it's a good time to host the loan related functionality in a self hosted module and then contribute to the existing modules separately. Development should be much more clear in that way.

harshjv:
@mattBlackDesign @monokh instead of hosting them separately, they can host it here and leverage our build system. If the changes are specifically made to their subpackages, we can let them merge it on their schedule and do a combine review when it is related to the common packages.

For now, the diff is very noisy because of unnecessary changes `liquality` to `atomicloans`. @mattBlackDesign can you first fix it? It will be easy to review then and we can quickly move forward.

monokh:
Going to be quite blunt here because I think this can prevent a great deal of headaches in the future. 

There is a clear separation of responsibility between the loan providers and the rest of the codebase. The current set of contributors shouldn't be concerned with maintaining or engaging in any real way with the loan related functionality. There are several areas where disagreements and conflict can and will occur in the future if this github project becomes a playground for every addition to the CAL:

**Code access writes**
Loans contributors shouldn't be required to go through approval to get code merged into the repo. Similarly, the other contributors shouldn't be distracted with changes that don't concern them, however small that may be. 

**Code responsibility**
Every independent module of CAL should be free to tag, release, branch on their own schedule. This one repo system not only accumulates a lot of unrelated components, it also slows down the concerning teams with delivering. 
What happens when a loan module fails the build? Who is responsible?
What happens when there is a new version release required immediately but core contributors are not there?
What happens when a non core contributor releases a version of the library which breaks certain things. Who is responsible?

**Infrastructure responsibility**
Infrastructure (building, deploying) has been setup on the repo for the project in a way that suits a given team. The build system and it's components (docs, releases etc.) may change with no notice to a party. At the same time, a party may want to customise these details, at which point, the infrastructure can become messy.

**Discussion and content**
Things like READMes, PRs, issues and github pages are unique content between loans and the rest of the project. Mixing it will make everything hard to find and organise.

matthewjablack:
@harshjv removed scope @atomicloans to reduce the number of diffs. This was in place temporarily to allow for those packages to be deployed to npm for atomic loans ui. 

You have both brought up very good points, and I agree that the loan functionality really has nothing to do with the rest of CAL. This could technically be a completely separate module, in a completely separate repo that is used alongside CAL, however we would need to determine how we want to setup the @liquality/client module to accommodate for that. 

We'll also need to think about how modules are published. I personally think that in the future all the modules will probably be their own repos, but I don't think we're at that stage yet.

packages/bitcoin-ledger-provider/lib/BitcoinLedgerProvider.js
```diff
@@ -176,7 +176,7 @@ export default class BitcoinLedgerProvider extends LedgerProvider {
       if (inputs && outputs) {
         let change = outputs.filter(output => output.id !== 'main')
 
-        if (change) {
+        if (change && change.length > 0) {
```
harshjv, line 179: Nice catch man!

Did you get any `absurdly-high-fee` error without this?

packages/client/lib/Chain.js
harshjv: @mattBlackDesign are you trying to get to `eth_pendingTransactions` or `parity_localTransactions` or `parity_pendingTransactions` via this method?

With above rpc calls, we can do something similar on eth side.

packages/client/lib/Chain.js
```diff
@@ -226,4 +234,16 @@ export default class Chain {
   async getAddressMempool (addresses) {
```
harshjv, line 234: Can you explain why `getAddressMempool`, `getCode` `erc20Approve`, `getTransactionReceipt` are needed on client.chain object?

harshjv (changes requested):

packages/bitcoin-ledger-provider/lib/BitcoinLedgerProvider.js
matthewjablack: Sorry for late reply @harshjv. 

Yes, in fact it wouldn't actually send the transaction because the fee was too high 😂

matthewjablack: Adding this to separate PR: https://github.com/liquality/chainabstractionlayer/pull/186

PR #185, opened 2019-06-04
Refund swaps

grandsmarquis:
### Description

This pull request introduce the possibility to monitor an existing swap and catch the refund action on all providers.

It also fixes some bugs inside of the Bitcoin Swap Js provider that were preventing party A to create a swap from the node.


packages/client/lib/Swap.js
```diff
@@ -158,4 +158,19 @@ export default class Swap {
 
     return this.client.getMethod('refundSwap')(initiationTxHash, recipientAddress, refundAddress, secretHash, expiration)
   }
+
+    /**
+   * Refund the swap
+   * @param {!string} initiationTxHash - The transaction hash of the swap initiation.
```
monokh, line 164: The rest of the params ?

packages/ethereum-swap-provider/lib/EthereumSwapProvider.js
```diff
@@ -176,6 +174,35 @@ export default class EthereumSwapProvider extends Provider {
...
+    let blockNumber = startBlock || await this.getMethod('getBlockHeight')()
+    const initiationTransaction = await this.getMethod('getTransactionReceipt')(initiationTxHash)
+    let refundSwapTransaction = false
+
+    while (!refundSwapTransaction) {
+      const block = await this.getMethod('getBlockByNumber')(blockNumber, true)
+      if (block) {
+        console.log(block.transactions)
+        for (const transaction of block.transactions) {
+          console.log( transaction.to === initiationTransaction.contractAddress,initiationTransaction.contractAddress, transaction.to)
```
monokh: guessing not needed?

test/integration/swap/common.js
```diff
@@ -132,6 +132,12 @@ async function refund (chain, initiationTxId, secretHash, swapParams) {
   return refundTxId
 }
 
+async function waitRefund(chain, initiationTxId, secretHash, swapParams) {
```
monokh: Can the `refund` function use the same model as `claimAndVerify`?

monokh (approved):
Great! just some minor comments

test/integration/swap/common.js
grandsmarquis: @monokh You mean passing the secret instead of secret hash bot for waitRefund() and refund() ?

monokh: I mean `refund` can be `refundAndVerify` which both submits the refund tx and tries to retrieve it using the find function. This way anywhere we do refund we are also testing that the find function works too

PR #187, opened 2019-06-04
Add sendBatchTransactions

matthewjablack:
### Description

Add extra outputs to send transaction ledger provider

This enables you to specify more than one output for a transaction in Bitcoin. 

For example: ` sendBatchTransaction ([{ to: address1, value: value1, data: script1 }, { to: address2, value: value2, data: script2 }])`

- [x] Implement sendBatchTransaction


packages/client/lib/Chain.js
```diff
@@ -197,8 +197,17 @@ export default class Chain {
    * @param {!string} from - The address from which the message is signed.
    * @return {Promise<string>} Resolves with a signed transaction.
    */
-  async sendTransaction (to, value, data, from) {
-    return this.client.getMethod('sendTransaction')(to, value, data, from)
+  async sendTransaction (to, value, data, from, extraOutputs) {
```
monokh: `extraOutputs` parameter should not be required anymore

packages/client/lib/Chain.js
```diff
@@ -197,8 +197,17 @@ export default class Chain {
...
+    return this.client.getMethod('sendTransaction')(to, value, data, from, extraOutputs)
+  }
+
+  /**
+   * Create, sign & broad a transaction with multiple outputs.
+   * @param {string[]} outputs - to, value, data, from.
+   * @return {Promise<string>} Resolves with a signed transaction.
+   */
+  async sendBatchTransaction (outputs) {
```
monokh: `outputs` is very bitcoin specific term. `transactions` would be more generic.

monokh (changes requested):

packages/client/lib/Chain.js
matthewjablack: good catch, removed 👍

packages/client/lib/Chain.js
matthewjablack: Good point. Changed `outputs` to `transactions`

PR #188, opened 2019-06-04
Add getCode and getTransactionReceipt to chain.js

matthewjablack:
### Description

Add `getCode` and `getTransactionReceipt` to `Chain.js`

- [x] Add `getCode`
- [x] Add `getTransactionReceipt` to `Chain.js`

### DIFF
--- a/packages/client/lib/Chain.js
+++ b/packages/client/lib/Chain.js
@@ -226,4 +226,12 @@ export default class Chain {
   async getAddressMempool (addresses) {
     return this.client.getMethod('getAddressMempool')(addresses)
   }
+
+  async getCode (address, block) {
+    return this.client.getMethod('getCode')(address, block)
+  }
+
+  async getTransactionReceipt (txHash) {
+    return this.client.getMethod('getTransactionReceipt')(txHash)
+  }
 }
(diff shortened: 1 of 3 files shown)


packages/client/lib/Chain.js
```diff
@@ -226,4 +226,12 @@ export default class Chain {
   async getAddressMempool (addresses) {
     return this.client.getMethod('getAddressMempool')(addresses)
   }
+
+  async getCode (address, block) {
```
monokh: The namespaces should only include functions that can be reasonably implemented cross chain. 
You can still access the `getCode` function without it being in the namespace by calling `client.getMethod('getCode')` or through calling the provider directly. 
It's fine to have it in the provider if you are building another provider that wants to access it.

monokh (changes requested):

packages/client/lib/Chain.js
matthewjablack: There could be the case for Bitcoin, where `getCode` is used to get the bitcoin script used to redeem a p2sh. Might be a bit far fetched though.

matthewjablack: I've removed this for now.

PR #189, opened 2019-06-05
Add additional ethereum utils

matthewjablack:
### Description

Add `removeAddress0x`, `checksumEncode` and `ensureBlockFormat` to Ethereum utils

`removeAddress0x` is for standardizing an Ethereum address and converting it to lower case

`checksumEncode` is for ensuring that an Ethereum address is following EIP 55

`ensureBlockFormat` is for ensuring that block parameter specified is using the correct format, when using calls such as `eth_call` (https://infura.io/docs/ethereum/json-rpc/eth_call). 

For example, you could have `client.chains.getCode(contractAddress, 'latest')` or  `client.chains.getCode(contractAddress, 7896092)` or even not include it, and it would ensure that it defaulted to `latest`. 

- [x] Add `removeAddress0x`
- [x] Add `checksumEncode`
- [x] Add `ensureBlockFormat`

### DIFF
--- a/packages/ethereum-utils/lib/index.js
+++ b/packages/ethereum-utils/lib/index.js
@@ -1,4 +1,6 @@
 import { Block, Transaction } from '@liquality/schema'
+import { padHexStart } from '@liquality/crypto'
+import eip55 from 'eip55'
 
 import { version } from '../package.json'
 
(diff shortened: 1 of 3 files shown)


monokh (approved):
This is fine as long as conflicts are fixed

PR #194, opened 2019-08-15
Segwit - Node & Ledger

monokh:
### Description

This PR implements segwit in the chainabstractionlayer. 

- [x] Implement segwit transaction support for waps
- [x] Implement segwit address support for ledger and node wallets

- [x] Remove bitcore provider
- [x] Remove JsLib provider - unify swap functionality into `BitcoinSwapProvider`
- [x] Move node wallet functionality into `BitcoinNodeWalletProvider`
- [x] Adds `BitcoinEsploraApiProvider` which covers the usecase of using your ledger to swap without requiring you running your own node.


packages/bitcoin-node-wallet-provider/lib/BitcoinNodeWalletProvider.js
```diff
@@ -0,0 +1,122 @@
...
+
+import { version } from '../package.json'
+
+const BIP70_CHAIN_TO_NETWORK = {
+  'main': BitcoinNetworks.bitcoin,
+  'test': BitcoinNetworks.bitcoin_testnet,
+  'regtest': BitcoinNetworks.bitcoin_regtest
+}
+
+const ADDRESS_TYPES = ['legacy', 'p2sh-segwit', 'bech32']
```
harshjv: Can we export this somewhere and reuse it? `BitcoinAddressTypes`?

packages/bitcoin-node-wallet-provider/lib/BitcoinNodeWalletProvider.js
```diff
@@ -0,0 +1,122 @@
...
+
+export default class BitcoinNodeWalletProvider extends WalletProvider {
+  constructor (network, uri, username, password, addressType = 'bech32') {
+    super()
+    if (!ADDRESS_TYPES.includes(addressType)) {
+      throw new Error(`addressType must be one of ${ADDRESS_TYPES.join(',')}`)
+    }
+    this.addressType = addressType
+    this.network = network
+    if (this.network.name === BitcoinNetworks.bitcoin.name) {
```
harshjv: Same with this logic. probably we can accommodate something like this in networks file. Just to reduce code duplication.

packages/bitcoin-rpc-provider/lib/BitcoinRpcProvider.js
```diff
@@ -35,45 +35,12 @@ export default class BitcoinRpcProvider extends JsonRpcProvider {
...
-    return this.jsonrpc('signmessage', from, message).then(result => Buffer.from(result, 'base64').toString('hex'))
-  }
-
   async sendTransaction (to, value) {
     to = addressToString(to)
     value = BigNumber(value).dividedBy(1e8).toNumber()
     return this.jsonrpc('sendtoaddress', to, value)
   }
 
-  async sendBatchTransaction (transactions) {
```
harshjv, line 49: I think @mattBlackDesign needed this for something.

packages/client/lib/Swap.js
```diff
@@ -42,6 +43,11 @@ export default class Swap {
    * @return {Promise<string>} Resolves with a 32 byte secret
    */
   async generateSecret (message) {
+    try {
+      return this.client.getMethod('generateSecret')(message)
```
harshjv, line 47: Are we implementing `generateSecret` anywhere else?

harshjv (commented):
utACK. @monokh can you remove `.DS_Store` and add it to `.gitignore` 😄

packages/bitcoin-rpc-provider/lib/BitcoinRpcProvider.js
gr0kchain: Yes, we should keep this

matthewjablack: Yep as discussed with @monokh the functionality of adding outputs to a `txb` in `bitcoinjs-lib` `TransactionBuilder` can be extracted to be used for both `sendTransaction` and `sendBatchTransaction`

monokh: It was just moved below the file. With the `BitcoinLedgerProvider`, it has been refactored to remove the duplication. 
Note that the functions used with the node provider to enable this are deprecated in the latest version of Bitcoin Core and will be removed in the next version.

packages/client/lib/Swap.js
monokh: Yes. `BitcoinNodeWalletProvider` has its own implementation because for node wallet, it is only possible to sign messages using a legacy address. `getAddresses` call returns addresses based on the type of address pre-specified in constructor (could be bech32/p2sh addresses) therefore for secret generation to continue working under these conditions, the node wallet needs to do it differently.

packages/bitcoin-node-wallet-provider/lib/BitcoinNodeWalletProvider.js
monokh: Bare with me on this one, actually our own custom `BitcoinNetworks` objects will be removed and the providers will use the BitcoinJsLib provided version. At which point, this mapping won't be required. Probably `BitcoinNetworks` will just alias to `bitcoinjs-lib.networks`

Just trying to get the main segwit part through, there are many improvements like this i am still pending to make

packages/bitcoin-node-wallet-provider/lib/BitcoinNodeWalletProvider.js
monokh: Moved to `bitcoin-utils`

packages/bitcoin-node-wallet-provider/lib/BitcoinNodeWalletProvider.js
monokh: Our custom network object now just aliases onto the BitcoinJsLib object with a few additional properties. 
Due to this, we no longer require 2 different objects to deal with this and this mapping is gone.

PR #213, opened 2019-09-18
Add feePerByte to sendTransaction

matthewjablack:
### Description

This PR adds `feePerByte` option to `sendTransaction`. 

- [x] Add `feePerByte` argument to `sendTranasction`, `sendBatchTransaction` for both `BitcoinJsWalletProvider` and `BitcoinLedgerProvider`


monokh:
I don't think the way this is done is right.
Anything specified in the normalised function signatures (i.e. `sendTransaction`) should be available on other chains. "feePerByte" is a bitcoin only concept. 
There are 2 valid ways of doing this. Either
a) include fee related parameters in the constructor - as done with the other bitcoin providers
or
b) develop a concept of fees and specifying them that is chain independent

PR #259, opened 2020-04-27
Fix eth findClaimSwapTransaction

harshjv:


monokh:
Can we also add for the error for tx receipt not found in refund while we are here. That scenario will not be hit, but it will be consistent

gr0kchain (approved):

PR #282, opened 2020-05-04
Fee support

monokh:
### Description

This PR adds fees functionality to the Chain Abstraction Layer. Allowing setting and updating transaction fees for regular and swap transactions.

- [x] Fee retrieval architecture
- [x] Fee setting and updating interface
- [x] Simple multiplier based fee providers
- [x] Time estimate in fee descriptor
- [x] Clean up modules (names, readmes)
- [x] Update docs (?)

BTC, Fee setting and updating
- [x] JS provider
- [x] Node provider
- [x] Ledger provider
- [x] Swap
- [x] Api fee provider
- [x] Fee property in transactions

ETH, Fee setting and updating
- [x] JS Provider
- [x] Node provider
- [x] Ledger provider
- [x] Metamask provider
- [x] Swap
- [x] EthGasStation fee provider
- [x] Fee property in transactions

ERC20
- [x] Swaps
- [x] Providers

Tests
- [x] Switch to using geth for integration tests (ganache does not support mempool https://github.com/trufflesuite/ganache-core/issues/484)
- [x] BTC send replace fee
- [x] ETH send replace fee
- [x] BTC Swap transactions fee
- [x] ETH Swap transactions fee

Extra: 
- [x] Fixed integration tests
- [x] Introduced testing environment for easy running integration tests locally and in build
- [x] Abstracted wallet aspects of `BitcoinLedgerProvider` and `BitcoinJsLibProvider` into `BitcoinWalletProvider`


packages/bitcoin-earn-fee-provider/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "type": "git",
+    "url": "git+https://github.com/liquality/chainabstractionlayer.git"
+  },
+  "bugs": {
+    "url": "https://github.com/liquality/chainabstractionlayer/issues"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "gitHead": "1855d9ec369c8c577f74eb04e310a7c00b9eaa6c"
```
harshjv, line 39: `gitHead` shouldn't be here. Did you abort version bump?

packages/bitcoin-esplora-api-provider/lib/BitcoinEsploraApiProvider.js
```diff
@@ -134,13 +134,18 @@ export default class BitcoinEsploraApiProvider extends Provider {
 
     const confirmations = tx.status.confirmed ? currentHeight - tx.status.block_height + 1 : 0
 
+    const totalFee = tx.fee
+    const fee = Math.round(totalFee / Math.ceil(tx.weight / 4))
```
harshjv, line 138: Can we use `BigNumber.js` here? Division operation may create issues.

packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.js
```diff
@@ -1,48 +1,34 @@
...
-
-export default class BitcoinJsWalletProvider extends Provider {
+export default class BitcoinJsWalletProvider extends BitcoinWalletProvider(WalletProvider) {
   constructor (network, mnemonic, addressType = 'bech32') {
-    super()
-    if (!AddressTypes.includes(addressType)) {
-      throw new Error(`addressType must be one of ${AddressTypes.join(',')}`)
-    }
+    super(network, addressType, [network])
     if (mnemonic === '') {
```
harshjv: Can we instead check for truthiness of `mnemonic`?

```
if (mnemonic) {}
```

packages/client/lib/Chain.js
```diff
@@ -19,29 +19,14 @@ export default class Chain {
   /**
    * Generate a block
    * @param {!number} numberOfBlocks - Number of blocks to be generated
-   * @return {Promise<string[], TypeError|InvalidProviderResponseError>} Resolves
-   *  with Block hash of the generated blocks.
-   *  Rejects with TypeError if input is invalid.
-   *  Rejects with InvalidProviderResponseError if provider's response is invalid.
+   * @return {Promise>}
```
harshjv: Typo here :)

packages/client/lib/Chain.js
```diff
@@ -19,29 +19,14 @@ export default class Chain {
...
-    }
-
-    const invalidBlock = find(blockHashes, blockHash => !(/^[A-Fa-f0-9]+$/.test(blockHash)))
-
-    if (invalidBlock) {
-      throw new InvalidProviderResponseError('Invalid block(s) found in provider\'s reponse')
-    }
-
-    return blockHashes
+    await this.client.getMethod('generateBlock')(numberOfBlocks)
```
harshjv: You can put `return` here instead of `await`.

packages/bitcoin-earn-fee-provider/package.json
monokh: I usually do not version bump until its in the branch (Dev). Does that cause an issue?

packages/client/lib/Chain.js
```diff
@@ -215,11 +200,29 @@ export default class Chain {
    * @param {!string} to - Recepient address.
    * @param {!string} value - Value of transaction.
    * @param {!string} data - Data to be passed to the transaction.
-   * @param {!string} from - The address from which the message is signed.
+   * @param {!string} [fee] - Fee price in native unit (e.g. sat/b, wei)
```
harshjv, line 203: I think we need `from` address for ETH.

harshjv (changes requested):

packages/client/lib/Chain.js
monokh: i didn't intend to return a response... but doesn't hurt i guess

packages/client/lib/Chain.js
monokh: fixed

packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.js
monokh: fixed

packages/bitcoin-esplora-api-provider/lib/BitcoinEsploraApiProvider.js
monokh: Doing rounding with bignumber is super annoying
and this shouldn't really cause an issue like this? value is in sats (well below max int limits) and since every value computed is intended for rounding anyway (.round, and .ceil), precision should not affect it

packages/client/lib/Chain.js
monokh: I checked to see if we have any uses for the from field and i couldn't find any. That's why i opted to replace the parameter so it's easier to use the fee (you don't always have to pass undefined for `from`)
Actually since this is in the abstraction layer interface, "from" is not really valid other than for ethereum
maybe you are thinking of multiple account support. that one actually needs thought how to add but in this interface, it would introduce ethereum conditional code into the apps.

PR #287, opened 2020-06-13
Fixed signMessage, added verifyMessage

gr0kchain:
### Description

This PR currently fixes an issue with calling the bitcoinmessage sign calls for supporting segwit addresses. Also added message verification


kraikov:
CAL migrated to Chainify

PR #335, opened 2020-08-19
Send return tx

monokh:
### Description

Changes send tx calls to return the transaction object that was sent

- [x] Update tx calls `sendTransaction` `initiateSwap` etc. to return tx object
- [x] Allow updateTransactionFee to take transaction object


harshjv (approved):
LGTM 🚀

PR #358, opened 2020-09-30
Add build and send sweep transaction for bitcoin js wallet

matthewjablack:
### Description

This PR adds `buildSweepTransaction` and `sendSweepTransaction` for Bitcoin js wallet enabling the ability to sweep all utxos to a destination of choice. 

This is particularly useful if you wish to remove all funds from a wallet

Note: This can be useful for Ethereum as well. The major use case would be for sweeping all ETH from a wallet (since it's already pretty easy to do this for ERC20's)

- [x] Add `buildSweepTransaction
- [x] Add `sendSweepTransaction
- [x] Add tests for `sendSweepTransaction`

### DIFF
--- a/packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.js
+++ b/packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.js
@@ -92,6 +92,29 @@ export default class BitcoinJsWalletProvider extends BitcoinWalletProvider(Walle
     return { hex: txb.build().toHex(), fee }
   }
 
+  async _buildSweepTransaction (externalChangeAddress, _outputs = [], feePerByte, fixedInputs) {
+    let _feePerByte = feePerByte || false
+    if (_feePerByte === false) _feePerByte = await this.getMethod('getFeePerByte')()
+
+    const { inputs, outputs, change } = await this.getInputsForSweep(_outputs, _feePerByte, fixedInputs)
+
+    if (change) {
+      throw Error('There should not be any change for sweeping transaction')
+    }
+
+    _outputs.forEach((output, i) => {
+      const spliceIndex = outputs.findIndex((sweepOutput) => output.value === sweepOutput.value)
+      outputs.splice(spliceIndex, spliceIndex + 1)
+    })
+
+    _outputs.push({
+      to: externalChangeAddress,
+      value: outputs[0].value
+    })
+
+    return this._buildTransaction(_outputs, feePerByte, inputs)
+  }
+
   async signP2SHTransaction (inputTxHex, tx, address, prevout, outputScript, lockTime = 0, segwit = false) {
     const wallet = await this.getWalletAddress(address)
     const keyPair = await this.keyPair(wallet.derivationPath)
--- a/packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.js
+++ b/packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.js
@@ -96,7 +96,7 @@ export default class BitcoinJsWalletProvider extends BitcoinWalletProvider(Walle
     let _feePerByte = feePerByte || false
     if (_feePerByte === false) _feePerByte = await this.getMethod('getFeePerByte')()
 
-    const { inputs, outputs, change } = await this.getInputsForSweep(_outputs, _feePerByte, fixedInputs)
+    const { inputs, outputs, change } = await this.getInputsForAmount(_outputs, _feePerByte, fixedInputs, 100, true)
 
     if (change) {
       throw Error('There should not be any change for sweeping transaction')
--- a/packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.js
+++ b/packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.js
@@ -92,7 +92,7 @@ export default class BitcoinJsWalletProvider extends BitcoinWalletProvider(Walle
     return { hex: txb.build().toHex(), fee }
   }
 
-  async _buildSweepTransaction (externalChangeAddress, _outputs = [], feePerByte, fixedInputs) {
+  async _buildSweepTransaction (externalChangeAddress, feePerByte, _outputs = [], fixedInputs) {
     let _feePerByte = feePerByte || false
     if (_feePerByte === false) _feePerByte = await this.getMethod('getFeePerByte')()
 
--- a/packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.js
+++ b/packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.js
@@ -92,25 +92,20 @@ export default class BitcoinJsWalletProvider extends BitcoinWalletProvider(Walle
     return { hex: txb.build().toHex(), fee }
   }
 
-  async _buildSweepTransaction (externalChangeAddress, feePerByte, _outputs = [], fixedInputs) {
+  async _buildSweepTransaction (externalChangeAddress, feePerByte) {
     let _feePerByte = feePerByte || false
     if (_feePerByte === false) _feePerByte = await this.getMethod('getFeePerByte')()
 
-    const { inputs, outputs, change } = await this.getInputsForAmount(_outputs, _feePerByte, fixedInputs, 100, true)
+    const { inputs, outputs, change } = await this.getInputsForAmount([], _feePerByte, [], 100, true)
 
     if (change) {
       throw Error('There should not be any change for sweeping transaction')
     }
 
-    _outputs.forEach((output, i) => {
-      const spliceIndex = outputs.findIndex((sweepOutput) => output.value === sweepOutput.value)
-      outputs.splice(spliceIndex, spliceIndex + 1)
-    })
-
-    _outputs.push({
+    const _outputs = [{
       to: externalChangeAddress,
       value: outputs[0].value
-    })
+    }]
 
     return this._buildTransaction(_outputs, feePerByte, inputs)
   }
--- a/packages/bitcoin-wallet-provider/lib/BitcoinWalletProvider.js
+++ b/packages/bitcoin-wallet-provider/lib/BitcoinWalletProvider.js
@@ -292,7 +292,7 @@ export default superclass => class BitcoinWalletProvider extends superclass {
 
       const outputBalance = _targets.reduce((a, b) => a + (b['value'] || 0), 0)
 
-      const amountToSend = utxoBalance - (feePerByte * (((_targets.length + 1) * 40) + (utxos.length * 155))) // todo better calculation
+      const amountToSend = utxoBalance - (feePerByte * (((_targets.length + 1) * 39) + (utxos.length * 153))) // todo better calculation
 
       let targets = _targets.map((target, i) => ({ id: 'main', value: target.value }))
       targets.push({ id: 'main', value: amountToSend - outputBalance })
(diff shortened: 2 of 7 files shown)


packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.js
```diff
@@ -92,6 +92,29 @@ export default class BitcoinJsWalletProvider extends BitcoinWalletProvider(Walle
...
   }
 
+  async _buildSweepTransaction (externalChangeAddress, _outputs = [], feePerByte, fixedInputs) {
+    let _feePerByte = feePerByte || false
+    if (_feePerByte === false) _feePerByte = await this.getMethod('getFeePerByte')()
+
+    const { inputs, outputs, change } = await this.getInputsForSweep(_outputs, _feePerByte, fixedInputs)
+
+    if (change) {
+      throw Error('There should not be any change for sweeping transaction')
```
harshjv, line 102: I think you missed a `new` here ;)

monokh:
Hey @matthewjablack thanks for the PR here, this is a very cool function
To keep the compatibility of the library uniform, we usually need any new functions to have implementations across ETH and ERC20 too, is this something you can add?

matthewjablack:
Hey @monokh we're pretty focused on Bitcoin development at the moment, so unfortunately the ETH side of things would be out of scope of us.

monokh:
@matthewjablack To ensure the library is predictable and fits the purpose of being an abstraction layer between the chains, we really need any functionality implemented to have a suitable interface and be implemented across the supported chains where possible.

Maintaining single chain based functionality adds overhead that the library is not suited towards. For example the current signature for `async sendSweepTransaction (externalChangeAddress, outputs, fee, fixedInputs)` is not compatible with the ETH chain and will need breaking changes when implemented cross chain.

So we ask for new features to be implemented cross chain because it:

- Forces the right interface to be created at the client level
- Ensures that the functionality is indeed able to be abstracted in a usable way
- Ensures that between the chains, the function is uniform and behaves in a predictable way
- Works as expected for the target users of the library (those looking to abstract cross chain)

matthewjablack:
Hey @monokh good point about `async sendSweepTransaction (externalChangeAddress, outputs, fee, fixedInputs)` not being compatible with the ETH chain. 

I've changed it to `async sendSweepTransaction (externalAddress, fee, outputs, fixedInputs)` and implemented it on the ETH chain

packages/bitcoin-wallet-provider/lib/BitcoinWalletProvider.js
```diff
@@ -226,6 +236,102 @@ export default superclass => class BitcoinWalletProvider extends superclass {
       .then(({ unusedAddress }) => unusedAddress[key])
   }
 
+  async getInputsForSweep (_targets = [], _feePerByte, fixedInputs = [], numAddressPerCall = 100) {
```
monokh: This looks like a lot of duplication from `getInputsForAmount`. Can the logic there be reused?

packages/bitcoin-wallet-provider/lib/BitcoinWalletProvider.js
```diff
@@ -226,6 +236,102 @@ export default superclass => class BitcoinWalletProvider extends superclass {
...
+
+      if (feePerByte === false) feePerByte = await feePerBytePromise
+      const minRelayFee = await this.getMethod('getMinRelayFee')()
+      if (feePerByte < minRelayFee) {
+        throw new Error(`Fee supplied (${feePerByte} sat/b) too low. Minimum relay fee is ${minRelayFee} sat/b`)
+      }
+
+      const outputBalance = _targets.reduce((a, b) => a + (b['value'] || 0), 0)
+
+      const amountToSend = utxoBalance - (feePerByte * (((_targets.length + 1) * 39) + (utxos.length * 153))) // todo better calculation
```
monokh: I guess this would depend on the type of tx. Can https://github.com/bitcoinjs/coinselect be used here in any way?

packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.js
```diff
@@ -92,6 +92,29 @@ export default class BitcoinJsWalletProvider extends BitcoinWalletProvider(Walle
...
+    }
+
+    _outputs.forEach((output, i) => {
```
monokh: What is going on here with the `_outputs` and `fixedInputs` ?
The common set of function across chains would be to send all value in the wallet to a single address.

packages/client/lib/Chain.js
```diff
@@ -187,6 +187,30 @@ export default class Chain {
...
 
+  /**
+   * Create & sign a sweep transaction
+   * @param {!string} externalAddress - External Change address.
+   * @param {string[]} outputs - to, value.
+   * @param {number} [fee] - Fee price in native unit (e.g. sat/b, wei)
+   * @param {string[]} fixedInputs - txid, vout, address, value, satoshis, derivationPath
+   * @return {Promise<string>} Resolves with a signed transaction object.
+   */
+  async buildSweepTransaction (externalAddress, fee, outputs, fixedInputs) {
```
monokh: If this is not something that is implemented cross chain, it shouldn't be included at the client level. In special situations where you need to access something on a specific chain, you can do `client.getMethod('buildSweepTransaction')` or access it directly on the provider object.

packages/client/lib/Chain.js
```diff
@@ -187,6 +187,30 @@ export default class Chain {
...
+
+  /**
+   * Create, sign & broadcast a sweep transaction.
+   * @param {!string} externalAddress - External Change address.
+   * @param {string[]} outputs - to, value.
+   * @param {number} [fee] - Fee price in native unit (e.g. sat/b, wei)
+   * @param {string[]} fixedInputs - txid, vout, address, value, satoshis, derivationPath
+   * @return {Promise<Transaction>} Resolves with a signed transaction.
+   */
+  async sendSweepTransaction (externalAddress, fee, outputs, fixedInputs) {
```
monokh: On the client, signatures should not include anything that we are not able to abstract across chains. Should be `sendSweepTransaction (address, fee)`

packages/ethereum-js-wallet-provider/lib/EthereumJsWalletProvider.js
```diff
@@ -105,6 +106,21 @@ export default class EthereumJsWalletProvider extends Provider {
     return normalizeTransactionObject(formatEthResponse(txData))
   }
 
+  async sendSweepTransaction (externalAddress, _gasPrice) {
+    const addresses = await this.getAddresses()
+
+    const balance = await this.getMethod('getBalance')(addresses)
+
+    const [ gasPrice ] = await Promise.all([
+      _gasPrice ? Promise.resolve(_gasPrice) : this.getMethod('getGasPrice')()
```
monokh, line 115: I think this comes back as wei? Later on in `sendTransaction` we pass this as the fee parameter (`gasPrice`) and on that level it expects it in gwei.

monokh (changes requested):
Please also include the ERC20 implementation. I believe it would be in `EthereumErc20Provider`

packages/bitcoin-wallet-provider/lib/BitcoinWalletProvider.js
matthewjablack: Tests passed with `legacy`, `p2sh-segwit` and `bech32`

We know that all utxos spent through this method, will be a p2pkh or p2wpkh txs (since we're spending from the wallet and not from p2sh or p2wsh txs) 

Looking at: https://github.com/bitcoinjs/coinselect/blob/master/utils.js#L3 we can see that

```
var TX_INPUT_BASE = 32 + 4 + 1 + 4
var TX_INPUT_PUBKEYHASH = 107
var TX_OUTPUT_BASE = 8 + 1
var TX_OUTPUT_PUBKEYHASH = 25
```

So a pkh input is (32 + 4 + 1 + 4) + 107 = 148
and a pkh output is (8 + 1) + 25 = 34

I added a bit of padding (5 bytes each) to arrive at 39 and 153 respectively

packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.js
matthewjablack: Having `_outputs` and `fixedInputs` allows you to spend all utxos in a wallet, and send a specific amount to an address, while sending the change to another address. 

Useful if you want to pay someone and sweep your wallet at the same time.

packages/ethereum-js-wallet-provider/lib/EthereumJsWalletProvider.js
matthewjablack: So `getGasPrice` function returns the value in `gwei` (tested locally with ganache, idk if it it's different with parity/geth/infura)

So I'm pretty sure gasPrice in `gwei` is being passed around (otherwise there would be issues here: https://github.com/liquality/chainabstractionlayer/blob/dev/packages/ethereum-js-wallet-provider/lib/EthereumJsWalletProvider.js#L96 )

packages/client/lib/Chain.js
matthewjablack: Good point, updated in https://github.com/liquality/chainabstractionlayer/pull/358/commits/3100c572312cdd9a3510e223d7ec76ca921a060e

packages/client/lib/Chain.js
matthewjablack: Good point, updated in https://github.com/liquality/chainabstractionlayer/pull/358/commits/3100c572312cdd9a3510e223d7ec76ca921a060e

packages/bitcoin-wallet-provider/lib/BitcoinWalletProvider.js
matthewjablack: DRYed this out by moving logic for `getInputsForSweep` into `getInputsForAmount` without affecting existing functionality of `getInputsForAmount` in https://github.com/liquality/chainabstractionlayer/pull/358/commits/68355ac3b6fc671b8630cd7f44415609f907d633

matthewjablack:
@monokh added `sendSweepTransaction` to `EthereumERC20Provider` and addressed all your comments

packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.js
monokh: Ok, i see what you are going for here. However i think the functions should not cater for complex features, unless it can be abstracted cross chain or is required for a specific need.

monokh: we want to reduce our maintenance to what is necessary for common usage

monokh: changing that should make this function as well as `getInputsForAmount` simpler

packages/bitcoin-wallet-provider/lib/BitcoinWalletProvider.js
monokh: This looks good. Unfortunately fees in `coinselect` are pre segwit, so likely not much we can do here for now: https://github.com/bitcoinjs/coinselect/issues/53

monokh (changes requested):
This is looking in really good shape! Just one final comment

packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.js
matthewjablack: There are actually a couple of use cases where this is quite important. 

If you want to allow users to select specific utxos and spend them or send funds to a particular location (similar to Wasabi) having defined inputs and outputs is necessary. 

Another case is for DLC's. Every DLC created, must commit to a specific utxo for the funding tx. If the user wants to create multiple DLC's, they must split up the utxos beforehand. To do this as efficiently as possible, it's important to be able to specify a particular utxo, create one output to be used for a specific DLC, and the other one be change.

matthewjablack:
Updated @monokh to remove Bitcoin specific functionality. This can be implemented externally if developers wish to allow wasabi type functionality

monokh (approved):
Looking good! Thanks for this

PR #363, opened 2020-10-13
WIP: Do not merge: Liquid Wallet Provider

tiero:
### Description

This PR serves as basic PoC of what a liquid wallet provider implementation could look like.

In order to test it, pull and checkout to current branch, enter into `packages/liquid-wallet-provider` and do `npm install`

Then go back to the root folder and run `npm run test:liquid`, it should print out 5 confidential blech32 addresses and 5 bech32 as well.

### DIFF
--- a/packages/liquid-wallet-provider/lib/LiquidWalletProvider.js
+++ b/packages/liquid-wallet-provider/lib/LiquidWalletProvider.js
@@ -1,0 +1,174 @@
+import * as liquid from 'liquidjs-lib'
+import * as slip77 from 'slip77'
+import * as bip32 from 'bip32'
+import { mnemonicToSeed } from 'bip39'
+
+
+import { LiquidAddressTypes } from './utils'
+import { Address, asyncSetImmediate } from '@liquality/utils'
+import { ConfidentialAddress } from './confidentialAddress'
+
+
+const ADDRESS_TYPE_TO_PREFIX = {
+  'legacy': 44,
+  'p2sh-segwit': 49,
+  'bech32': 84,
+  'legacy-confidential': 44,
+  'p2sh-segwit-confidential': 49,
+  'blech32': 84
+}
+
+export default class LiquidWalletProvider {
+
+  constructor(network, mnemonic, addressType = 'blech32', superArgs = []) {
+    if (!LiquidAddressTypes.includes(addressType)) {
+      throw new Error(`addressType must be one of ${LiquidAddressTypes.join(',')}`)
+    }
+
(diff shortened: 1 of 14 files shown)


packages/liquid-wallet-provider/lib/LiquidWalletProvider.js
```diff
@@ -0,0 +1,174 @@
...
+  'p2sh-segwit': 49,
+  'bech32': 84,
+  'legacy-confidential': 44,
+  'p2sh-segwit-confidential': 49,
+  'blech32': 84
+}
+
+export default class LiquidWalletProvider {
+
+  constructor(network, mnemonic, addressType = 'blech32', superArgs = []) {
```
monokh, line 23: Is a liquid wallet backwards compatible with regular bitcoin wallets? If so, do you think it would make sense that this class actually extends `BitcoinWalletProvider`?

@harshjv  thoughts?

packages/liquid-wallet-provider/lib/confidentialAddress.js
```diff
@@ -0,0 +1,107 @@
+import { version } from '../package.json'
+
+function sleep(ms) {
+  return new Promise(resolve => setTimeout(resolve, ms))
+}
+
+class ConfidentialAddress {
+  constructor(address, derivationPath, publicKey, index, blindingKey) {
```
monokh, line 8: You could extend the original class to add the `blindingKey`

packages/liquid-wallet-provider/lib/LiquidWalletProvider.js
tiero: The address encoding differs (ie different prefixes and so on) and we have a new subset of addresses  called "confidential addresses" which are like bitcoin (legacy, wrapped segwit, native segwit) with the difference that a blinding public key is encoded inside as well, make them longer than unconfidential one. (ie. bech32 becomes blech32)

I don't think Confidnetial Transactions will ever merged to Bitcoin, too risky. So I would consider it as a different chain/address type.

packages/liquid-wallet-provider/lib/LiquidWalletProvider.js
```diff
@@ -0,0 +1,174 @@
...
+const ADDRESS_TYPE_TO_PREFIX = {
+  'legacy': 44,
+  'p2sh-segwit': 49,
+  'bech32': 84,
+  'legacy-confidential': 44,
+  'p2sh-segwit-confidential': 49,
+  'blech32': 84
+}
+
+export default class LiquidWalletProvider {
```
monokh, line 21: Should extend WalletProvider

PR #364, opened 2020-10-15
Add Liquid Provider

tiero:
# Description

This PR adds a LiquidWalletProvider and a LiquidJsWalletProvider using liquidjs-lib.

It introduces `Pset` (the Psbt for Elements chains)  interface. It is not a complete Liqiuid integrations since EsploraApi and Elements node RPC provider are missing. 

NOTICE: I'm not able to use `lerna bootstrap` since on Linux there is seems to be some problems, I need to investigate. For this reason circular dependency assume a relative path to get the dependency (ie. will break if package are installed by themself). i need help and guidance on that.

To test `npm run test:liquid`

- [x] LiquidWalletProvider
- [x] LiquidJsWalletProvider
- [ ] LiquidWalletProvider.sendTransaction
- [ ] LiquidEsploraApiProvider
- [ ] LiquidRpcProvider


kraikov:
CAL migrated to Chainify

PR #390, opened 2021-03-08
Use scraper's new events api

harshjv:


packages/ethereum-scraper-swap-find-provider/lib/EthereumScraperSwapFindProvider.js
```diff
@@ -68,21 +104,18 @@ export default class EthereumScraperSwapFindProvider extends NodeProvider {
...
     if (!initiationTransactionReceipt) throw new PendingTxError(`Transaction receipt is not available: ${initiationTxHash}`)
 
-    const transaction = await this.findAddressTransaction(initiationTransactionReceipt.contractAddress,
-      tx => this.getMethod('doesTransactionMatchClaim', false)(tx, initiationTransactionReceipt))
+    const transaction = await this.findAddressEvents('swapClaim', initiationTransactionReceipt.contractAddress,
+      tx => this.getMethod('doesTransactionMatchClaim')(tx, initiationTransactionReceipt))
     if (!transaction) return
 
-    if (transaction._raw.status === true) {
-      transaction.secret = await this.getMethod('getSwapSecret')(transaction.hash)
```
monokh, line 76: I see the secret is not appended anymore
Does this come in the transaction from `findAddressEvents`? I think normalizeTransactionObject will strip it and it will end up in `tx._raw.secret` but not `tx.secret`. The clients rely on this path

PR #396, opened 2021-04-14
Near typescript package integration in chain abstraction layer

tima-t:


packages/near-rpc-provider/lib/NearRpcProvider.ts
```diff
@@ -0,0 +1,185 @@
...
+    const blockNumber = await this.getBlockHeight(tx.transaction_outcome.block_hash)
+    return { ...tx, blockNumber }
+  }
+
+  async getGasPrice() {
+    const result = await this._rpcQuery('gas_price', [null])
+    return get(result, 'gas_price')
+  }
+
+  async getBalance(addresses: string[]): Promise<BigNumber> {
```
monokh: `getBalance` now takes `(addresses: (Address | string)[])`

packages/near-rpc-provider/lib/NearRpcProvider.ts
```diff
@@ -0,0 +1,185 @@
...
+    try {
+      if (!isArray(addresses)) {
```
monokh: this is not needed. always should be array

packages/near-rpc-provider/lib/NearRpcProvider.ts
```diff
@@ -0,0 +1,185 @@
...
+      return new BigNumber(balance.available)
+    } catch (err) {
+      if (err.message && err.message.includes('does not exist while viewing')) {
+        return new BigNumber(0)
+      }
+      throw err
+    }
+  }
+
+  async isAddressUsed(address: string): Promise<boolean> {
```
monokh: This method no longer exists on the other providers, we never used it so was removed in the latest version

packages/near-swap-find-provider/lib/NearSwapFindProvider.ts
```diff
@@ -0,0 +1,176 @@
...
+      const normalizedTransactions = {} as { [key: string]: near.NearSwapTransaction }
+
+      for (const tx of transactions) {
+        normalizedTransactions[tx.hash] = {
+          ...normalizedTransactions[tx.hash],
+          ...this.normalizeTransactionResponse(tx)
+        }
+      }
+
+      const tx = Object.values(normalizedTransactions).find(predicate) as near.NearSwapTransaction
```
monokh: Hmm why does it need to be cast here? It should be that type

packages/near-swap-find-provider/lib/NearSwapFindProvider.ts
```diff
@@ -0,0 +1,176 @@
...
+
+      if (tx) {
+        const currentHeight = await this.getMethod('getBlockHeight')()
+        const txBlockHeight = await this.getMethod('getBlockHeight')(tx.blockHash)
+        tx.confirmations = currentHeight - txBlockHeight
+        return { ...tx, secret: tx.swap.secret, _raw: tx } as Transaction<near.NearSwapTransaction>
```
monokh: Might be wrong here, but doesn't casting like this with `as` cause the type checking to be ignored? 
i'm worried some fields will be incorrect by doing it like this

packages/near-rpc-provider/lib/NearRpcProvider.ts
kraikov: Should be fine now. I've also improved the way balance is calculated.

packages/near-rpc-provider/lib/NearRpcProvider.ts
kraikov: Thanks. Should be fixed now.

packages/near-rpc-provider/lib/NearRpcProvider.ts
kraikov: Thanks, wasn't aware of that. It's now removed.

packages/near-swap-find-provider/lib/NearSwapFindProvider.ts
kraikov: Yeah, that casting was redundant.

packages/near-swap-find-provider/lib/NearSwapFindProvider.ts
kraikov: Yeah, that casting was redundant.

PR #415, opened 2021-05-12
Near swap validation

kraikov:
### Description

- remove `validateSwapParams` from claimSwap and `refundSwap` because it's called inside `verifyInitiateSwapTransaction` 
- add swap params validation for Near

### DIFF
--- a/packages/near-swap-find-provider/lib/NearSwapFindProvider.ts
+++ b/packages/near-swap-find-provider/lib/NearSwapFindProvider.ts
@@ -2,7 +2,7 @@ import { near, SwapParams, SwapProvider, Transaction } from '@liquality/types'
 import { NodeProvider } from '@liquality/node-provider'
 import { PendingTxError } from '@liquality/errors'
 import { addressToString } from '@liquality/utils'
-import { fromBase64, toBase64, fromNearTimestamp, parseReceipt } from '@liquality/near-utils'
+import { fromBase64, toBase64, fromNearTimestamp, parseReceipt, validateSwapParams } from '@liquality/near-utils'
 
 const ONE_DAY_IN_NS = 24 * 60 * 60 * 1000 * 1000 * 1000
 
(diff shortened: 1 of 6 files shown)


packages/near-utils/lib/index.ts
```diff
@@ -1,9 +1,18 @@
...
 import BN from 'bn.js'
 
+export { validateSecret, validateSecretAndHash } from '@liquality/utils'
 export { transactions, Account, InMemorySigner, providers, KeyPair, keyStores } from 'near-api-js'
 export { BN }
 
+export function validateSwapParams(swapParams: SwapParams) {
+  validateValue(swapParams.value)
+  validateSecretHash(swapParams.secretHash)
+  validateExpiration(swapParams.expiration)
```
monokh, line 15: can there also be some reasonable validation for the address? Such that it is a string and within limits? Here's some paramters: https://nomicon.io/DataStructures/Account.html

packages/near-swap-find-provider/lib/NearSwapFindProvider.ts
```diff
@@ -134,8 +136,9 @@ export default class NearSwapFindProvider extends NodeProvider implements Partia
     swapParams: SwapParams,
     initiationTxHash: string
   ): Promise<Transaction<near.NearSwapTransaction>> {
-    const initiationTransactionReceipt = await this.getMethod('getTransactionReceipt')(initiationTxHash)
+    validateSwapParams(swapParams)
```
monokh, line 147: before returning the tx, also secret needs to be validated like 
`validateSecretAndHash(tx.secret, swapParams.secretHash)`

monokh (commented):
Could you add some tests for the validation to `near-utils`? You can copy most from  `ethereum-swap-provider` tests

Also do you feel it's better to just stick the `NearSwapFindProvider` into `NearSwapProvider` ? It seems they cannot be separated. That will make it simpler

monokh (approved):

PR #416, opened 2021-05-14
Bitcoin Cash

DesWurstes:
As discussed


packages/bitcoin-cash-rest-api-provider/lib/BitcoinCashRestApiProvider.ts
```diff
@@ -0,0 +1,197 @@
...
+      status.block_height = currentHeight - rawTx['confirmations'] + 1
+      status.block_hash = await this.getBlockHash(status.block_height)
+    }
+
+    const tx: explorer.Transaction = {
+      txid: rawTx['txid'],
+      version: rawTx['version'],
+      locktime: rawTx['locktime'],
+      vin: rawTx['vin'],
+      fee: 0, // TODO
```
DesWurstes, line 109: This is not a "will do" but rather a "don't care" that I justified in the last section of [this](https://github.com/liquality/chainabstractionlayer/pull/405#issue-630515718). Simply, costs outweigh no use for this number.

packages/bitcoin-cash-rest-api-provider/lib/BitcoinCashRestApiProvider.ts
```diff
@@ -0,0 +1,197 @@
+import { NodeProvider } from '@liquality/node-provider'
+import { addressToString } from '@liquality/utils'
+import { decodeRawTransaction, normalizeTransactionObject } from '@liquality/bitcoin-utils'
+import { TxNotFoundError, BlockNotFoundError } from '@liquality/errors'
+import { ChainProvider, Address, bitcoin, BigNumber } from '@liquality/types'
+import * as explorer from './types'
+import { BitcoinCashNetwork } from '@liquality/bitcoin-networks'
+import BCHJS from '@psf/bch-js'
```
monokh: If I understand correctly, this package simply makes rest calls. 
Throughout the library, we have avoided using rest clients to avoid introducing an extra dependency (security).
Any chance these calls can be made with axios?

packages/bitcoin-cash-rest-api-provider/lib/BitcoinCashRestApiProvider.ts
```diff
@@ -0,0 +1,197 @@
...
+
+    this._network = network
+    this._numberOfBlockConfirmation = numberOfBlockConfirmation
+    this._defaultFeePerByte = defaultFeePerByte
+    this._usedAddressCache = {}
+    this.bchjs = new BCHJS({ restURL: url })
+  }
+
+  async getFeePerByte(numberOfBlocks = this._numberOfBlockConfirmation) {
+    numberOfBlocks as any
```
monokh: What does this line do?

packages/bitcoin-cash-rest-api-provider/lib/BitcoinCashRestApiProvider.ts
```diff
@@ -0,0 +1,197 @@
...
+      txid: utxo['tx_hash'],
+      vout: utxo['tx_pos'],
+      value: utxo['value'],
+      blockHeight: utxo['height']
+    }))
+  }
+
+  async getUnspentTransactions(_addresses: (Address | string)[]): Promise<bitcoin.UTXO[]> {
+    const addresses = _addresses.map(addressToString)
+    const utxoSets = await Promise.all(addresses.map((addr) => this._getUnspentTransactions(addr)))
```
monokh: This fullstack api [seems to support querying multiple addresses](https://api.fullstack.cash/docs/#api-ElectrumX_/_Fulcrum-UTXOs_for_an_array_of_addresses) in one query. Could we use that for the efficiency gain? We do similarly on the bitcoin side

packages/bitcoin-cash-rest-api-provider/lib/BitcoinCashRestApiProvider.ts
```diff
@@ -0,0 +1,197 @@
...
+    const utxos = flatten(utxoSets)
+    return utxos
+  }
+
+  async _getAddressTransactionCount(address: string) {
+    const transactions = await this.bchjs.Electrumx.transactions(address)
+    return transactions['transactions'].length
+  }
+
+  async getAddressTransactionCounts(_addresses: (Address | string)[]) {
```
monokh, line 78: Any batch api available here similar to the utxo one?

packages/bitcoin-cash-rest-api-provider/lib/types.ts
```diff
@@ -0,0 +1,78 @@
+export type FeeEstimates = { [index: string]: number }
```
monokh: This seems identical to the `BitcoinEsploraApiProvider` types
In that case it should be moved to the types package, in the bitcoin folder i guess

packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.ts
```diff
@@ -26,7 +34,11 @@ export default class BitcoinJsWalletProvider extends BitcoinWalletProvider(
 
   constructor(options: BitcoinJsWalletProviderOptions) {
     const { network, mnemonic, baseDerivationPath, addressType = bitcoin.AddressType.BECH32 } = options
-    super({ network, baseDerivationPath, addressType })
+    super({
+      network,
+      baseDerivationPath,
+      addressType: network.segwitCapable ? addressType : bitcoin.AddressType.LEGACY
```
monokh: This is making it a bit unclear IMO, if the parameter passed is incorrect then it shouldn't be silently changed

An alternative can be to throw an error if `segwitCapable` is false and the wrong address type is passed

Users of the library should pass legacy when using BCH

packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.ts
```diff
@@ -156,22 +208,59 @@ export default class BitcoinJsWalletProvider extends BitcoinWalletProvider(
...
 
     const sigs = []
     for (let i = 0; i < inputs.length; i++) {
       const index = inputs[i].txInputIndex ? inputs[i].txInputIndex : inputs[i].index
+
+      if (this._network.protocolType == ProtocolType.BitcoinCash) {
+        const sigHash = tx.hashForWitnessV0(index, inputs[i].outputScript, inputs[i].vout.vSat, 0x01 | 0x40)
+        const signed = keyPairs[i].sign(sigHash)
+
+        // BitcoinJS does not allow SIGHASH_FORKID for encoding
```
monokh: Could you potentially extract this function and add some comments about what it does, perhaps a source for the code?

packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.ts
```diff
@@ -156,22 +208,59 @@ export default class BitcoinJsWalletProvider extends BitcoinWalletProvider(
...
+      const wallet = await this.getWalletAddress(addr)
       const keyPair = await this.keyPair(wallet.derivationPath)
       keyPairs.push(keyPair)
     }
 
     const sigs = []
     for (let i = 0; i < inputs.length; i++) {
       const index = inputs[i].txInputIndex ? inputs[i].txInputIndex : inputs[i].index
+
+      if (this._network.protocolType == ProtocolType.BitcoinCash) {
```
monokh: Is it specific to bitcoin cash or is there more generality to it? If other forks have the same process, should call the parameter `ProtocolType.BitcoinCash` something more generic

monokh: not sure if this sounds right, but maybe there can be a `forkId` parameter in the network ?

packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.ts
```diff
@@ -64,6 +76,41 @@ export default class BitcoinJsWalletProvider extends BitcoinWalletProvider(
...
+      vout.vSat = inputs[i].value
+      inputInfo.push({
+        inputTxHex: '',
+        index: i,
+        vout: vout,
+        outputScript: address.toOutputScript(this._network.addrToBitcoinJS(inputs[i].address), this._network as any)
+      })
+      inputAddressInfo.push(inputs[i].address)
+    }
+    const sigs = await this.signBatchP2SHTransaction(inputInfo as any, inputAddressInfo as any, tx, 0)
```
monokh: Why the `as any`? Parameters seem compatible

packages/bitcoin-networks/lib/index.ts
```diff
@@ -30,4 +70,158 @@ const BitcoinNetworks = {
...
+      return transformer(address)
+    } catch (e) {
+      return address
+    }
+  }
+}
+
+// Bitcoin Cash
+
+function cashaddrToLegacy(address: string) {
```
monokh: We like to keep the BitcoinNetwork class serializable as it gets passed between contexts sometimes

Is this format only for bitcoin cash? In that case it would be great if it could be moved to `bitcoin-utils` instead. 

something like 

```
addToBitcoinJs (address: string, network: BitcoinNetwork) {
 if(!bitcoincash) return address

  ....
 // handle the cash address format
}
```

packages/bitcoin-swap-provider/lib/BitcoinSwapProvider.ts
```diff
@@ -37,7 +37,7 @@ export default class BitcoinSwapProvider extends Provider implements Partial<Swa
       throw new Error(`Mode must be one of ${swapModes.join(',')}`)
     }
     this._network = network
-    this._mode = mode
+    this._mode = network.segwitCapable ? mode : bitcoin.SwapMode.P2SH
```
monokh: Similar to the other point about throwing an error if the wrong type was supplied

packages/bitcoin-cash-rest-api-provider/lib/BitcoinCashRestApiProvider.ts
DesWurstes: Used to silence a warning

DesWurstes:
> This seems identical to the BitcoinEsploraApiProvider types
In that case it should be moved to the types package, in the bitcoin folder i guess

Fixed, moved it to types. Please make sure to update calling CAL from other libs because this is a breaking change: esplora-api-provider no longer exports esplora, and it's..... **Never mind I moved it now in a way that the previous API still exists in a working condition**

> Is it specific to bitcoin cash or is there more generality to it? If other forks have the same process, should call the parameter ProtocolType.BitcoinCash something more generic
@monokh monokh 20 hours ago Member

> not sure if this sounds right, but maybe there can be a forkId parameter in the network ?

This is specific to BCH, BCHABC, BCHSV and BTG. BTG does not even sign the same way BCH does, while BCHABC and BCHSV both use the same forkId as BCH, so I don't see it as worth implementing. BCHABC, BCHSV fit perfectly in the current BitcoinCashNetwork struct, but if BTG were to be added, then it would be closer to BitcoinNetwork which would need forkId. Simply, I don't think we should add that.

I have nothing to add, but further reviews and ideas still welcome :)

packages/bitcoin-esplora-swap-find-provider/lib/BitcoinEsploraSwapFindProvider.ts
```diff
@@ -10,18 +10,28 @@ type PaymentVariants = {
 }
 
 export default class BitcoinEsploraSwapFindProvider extends NodeProvider {
-  constructor(url: string) {
+  _useRest: boolean
```
monokh: it looks like you are trying to support 2 apis with the same code, technically all the requests are rest calls
would it be better to create another provider that extends this and implements the `findAddressTransaction` differently?

DesWurstes:
This batch provider turned out to need to use the Ninsight API. The BCHJS package that used to be there was actually hiding that this was using a separate API. Done.

monokh (approved):

monokh (approved):

PR #417, opened 2021-05-18
added polygon in ethereum-networks

tima-t:
### Description

Add polygon config in ethereum-networks

### DIFF
--- a/packages/ethereum-networks/lib/index.ts
+++ b/packages/ethereum-networks/lib/index.ts
@@ -94,7 +94,7 @@ const bsc_testnet: EthereumNetwork = {
 }
 
 const polygon_mainnet: EthereumNetwork = {
-  name: 'matic_mainnet',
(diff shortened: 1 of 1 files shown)


packages/ethereum-networks/lib/index.ts
```diff
@@ -93,6 +93,22 @@ const bsc_testnet: EthereumNetwork = {
   isTestnet: true
 }
 
+const polygon_mainnet: EthereumNetwork = {
+  name: 'matic_mainnet',
```
monokh: why the difference between the name and the variable name?

monokh (changes requested):

packages/ethereum-networks/lib/index.ts
tima-t: This one is fixed, the name should be polygon_mainnet.

monokh (approved):

PR #432, opened 2021-07-09
Feat/solana integration

kraikov:
### Description
Solana packages.


packages/solana-network/lib/index.ts
```diff
@@ -0,0 +1,35 @@
+import { Network } from '@liquality/types'
```
monokh: Package name should be consistent with others i.e. `solana-networks` vs `solana-network`

packages/solana-rpc-provider/lib/SolanaRpcProvider.ts
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  async getTransactionReceipt(txHashes: string[]): Promise<Transaction<solana.InputTransaction>[]> {
+    const transactions = await this.connection.getParsedConfirmedTransactions(txHashes)
+
+    if (!transactions) {
+      return []
+    }
+
+    return transactions.map((transaction) => normalizeTransaction(transaction))
```
monokh, line 90: I don't think this one needs to be normalized? I believe it's only used internally in the library and not exposed outside. I think it will make it more simpler in the code, instead of trying to access from`_raw` all the time

packages/solana-rpc-provider/lib/SolanaRpcProvider.ts
```diff
@@ -0,0 +1,175 @@
...
+  Keypair,
+  BpfLoader,
+  BPF_LOADER_PROGRAM_ID,
+  ConfirmedSignatureInfo,
+  Signer
+} from '@solana/web3.js'
+
+import { addressToString } from '@liquality/utils'
+
+export default class SolanaRpcProvider extends NodeProvider implements Partial<ChainProvider> {
```
monokh: should also be implementing `FeeProvider` ?

packages/solana-swap-provider/lib/SolanaSwapProvider.ts
```diff
@@ -0,0 +1,205 @@
...
+  createClaimBuffer,
+  createInitBuffer,
+  createRefundBuffer,
+  validateSwapParams
+} from '@liquality/solana-utils'
+import { WalletError, TxNotFoundError, InvalidAddressError } from '@liquality/errors'
+
+export default class SolanaSwapProvider extends Provider implements Partial<SwapProvider> {
+  doesBlockScan: boolean | (() => boolean)
+  signer: Keypair
```
monokh: These are unused>

packages/solana-swap-provider/lib/SolanaSwapProvider.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  generateSecret(message: string): Promise<string> {
+    return sha256(message)
```
monokh: This will be unsafe, because it's deterministic and anyone can produce it. 
We use a standard message to generate the secret. So if you can just `sha256` the message for the secret, you can find the secret without the HTLC revealing it. In other chains we sign a message as that will produce a signature which only the generator of the secret can produce. 

Is there some way of doing the same with the solana keys?

packages/solana-swap-provider/lib/SolanaSwapProvider.ts
```diff
@@ -0,0 +1,205 @@
...
+
+    return transactionByHash?.secret
+  }
+
+  async initiateSwap(swapParams: SwapParams): Promise<Transaction> {
+    validateSwapParams(swapParams)
+
+    const signer = await this.getMethod('_getSigner')()
+
+    const { programId } = await this.getMethod('getConnectedNetwork')()
```
monokh: Rather explicitly pass the network into the provider constructor, this is how other swap providers are setup.

packages/solana-utils/lib/index.ts
```diff
@@ -0,0 +1,209 @@
...
+      }
+    }
+  }
+
+  if (!transactionData.programId) {
+    transactionData.programId = accountKeys[accountKeys.length - 1].pubkey.toString()
+  }
+
+  if (signatureStatus?.value?.confirmationStatus === 'finalized') {
+    transactionData.confirmations = 31
```
monokh, line 177: can you explain this magic number `31`?

packages/solana-wallet-provider/lib/SolanaWalletProvider.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  async getAddresses(): Promise<Address[]> {
+    if (this._addressCache[this._mnemonic]) {
+      return [this._addressCache[this._mnemonic]]
+    }
+
+    const account = await this._getSigner()
+
+    const result = new Address({
+      address: account.publicKey.toString(),
```
monokh, line 47: if public key is available, can also add to the Address object. It's fine if it's the same as address

test/integration/common.ts
```diff
@@ -316,6 +335,32 @@ async function fundAddress(chain: Chain, address: string, value?: BigNumber): Pr
...
+
+      solana.addProvider(new SolanaRpcProvider(config.solana.network))
+      solana.addProvider(
+        new SolanaWalletProvider({
+          network: config.solana.network,
+          mnemonic: config.solana.senderMnemonic,
+          derivationPath: `m/44'/501'/${config.solana.walletIndex}'/0'`
+        })
+      )
+      solana.addProvider(new SolanaSwapProvider())
```
monokh: you shouldn't need the swap providers to simply send some funds I imagine?

packages/solana-utils/lib/index.ts
```diff
@@ -0,0 +1,209 @@
...
+    secret_hash,
+    expiration,
+    value
+  })
+
+  return serialize(initSchema, initTemplate)
+}
+
+export function validateSecret(swapParams: SwapParams, data: { secret: string }): boolean {
+  _validateSecret(data.secret)
```
monokh: I think this should also validate secret and hash?
since we have the swap params available here, we can ensure that the secret matches the secret hash of the swap params. Using `validateSecretAndHash`

packages/solana-swap-find-provider/lib/SolanaSwapFindProvider.ts
```diff
@@ -0,0 +1,136 @@
...
+  private instructions = {
+    init: 0,
+    claim: 1,
+    refund: 2
+  }
+
+  async findInitiateSwapTransaction(swapParams: SwapParams): Promise<Transaction> {
+    validateSwapParams(swapParams)
+
+    const { programId } = await this.getMethod('getConnectedNetwork')()
```
monokh: better to explicitly pass the network

monokh (changes requested):

packages/solana-utils/lib/index.ts
AndonMitev: So transactions have own statuses depending on the number of the confirmations, once when they reach state finalized it means there are 31 confirmations on the chain

test/integration/common.ts
AndonMitev: nah i just follow the rest of the packages implemented in that scope, should I remove it or keep it for consistency like the other packages are?

packages/solana-rpc-provider/lib/SolanaRpcProvider.ts
AndonMitev: It does, this is the single point where transactions are fetched and parsed, this is used in swap provider, swap find provider etc so this point is very important

packages/solana-utils/lib/index.ts
monokh: ok, then can we use a combination of a constant and some comment to make sure it can be understable when reading the code?

monokh (approved):

PR #437, opened 2021-07-16
Feat/Terra Integration

AndonMitev:
:warning: THIS IS A PULL REQUEST TEMPLATE! :warning: This helps you complete a pull request. Please follow the instructions below. Reminder that you are bound to abide by the [Code of Conduct](./CODE_OF_CONDUCT.md)

### Description

Describe the purpose of this PR

- [ ] Add x
- [ ] Fix y
- [ ] Check z


packages/terra-swap-provider/lib/TerraSwapProvider.ts
```diff
@@ -0,0 +1,89 @@
...
+      messages: [initContractMsg]
+    })
+  }
+
+  async claimSwap(
+    swapParams: SwapParams,
+    initiationTxHash: string,
+    secret: string
+  ): Promise<Transaction<terra.InputTransaction>> {
+    validateSwapParams(swapParams)
```
kraikov: here we should also use `validateSecret` and `validateSecretAndHash`

packages/terra-swap-provider/lib/TerraSwapProvider.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    return transaction
+  }
+
+  async fundSwap(): Promise<null> {
+    return null
+  }
+
+  async verifyInitiateSwapTransaction(swapParams: SwapParams, initiationTxHash: string): Promise<boolean> {
+    const initTx = await this.getMethod('getTransactionByHash')(initiationTxHash)
```
kraikov, line 87: We should use `validateSwapParams(swapParams)` before making any calls

kraikov: when you have `validateSwapParams` inside `verifyInitiateSwapTransaction` you'll no longer need to call it inside claim and refund

packages/terra-utils/lib/index.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export const validateSwapParams = (swapParams: SwapParams) => {
+  validateValue(swapParams.value)
+  validateSecretHash(swapParams.secretHash)
+  validateExpiration(swapParams.expiration)
+  validateAddress(addressToString(swapParams.recipientAddress))
+  validateAddress(addressToString(swapParams.refundAddress))
+}
+
+const convertDateToTimestamp = (fullDate: string): number => {
```
kraikov, line 83: Can we try to use `luxon` or `dayjs` for that?

packages/terra-wallet-provider/lib/TerraWalletProvider.ts
```diff
@@ -0,0 +1,154 @@
...
+      msgs.push(this._sendMessage(to, value))
+    }
+
+    const tx = await wallet.createAndSignTx({
+      msgs
+    })
+
+    const transaction = await this.getMethod('_broadcastTx')(tx)
+
+    console.log(transaction.txhash)
```
kraikov: Remove.

packages/terra-utils/lib/index.ts
```diff
@@ -0,0 +1,87 @@
...
+export const normalizeBlock = (data: any): Block => ({
+  hash: data.block_id.hash,
+  timestamp: convertDateToTimestamp(data.block.header.time),
+  size: Number(data.block.header.height),
+  number: Number(data.block.header.height),
+  parentHash: data.block.last_commit.block_id.hash
+})
+
+export const normalizeTransaction = (data: any, asset: string): Transaction<terra.InputTransaction> => {
+  const value = data.tx?.msg?.[0]?.init_coins?.get(asset)?.amount || 0
```
kraikov: Just for the sake of readability can we use `get` from lodash?

packages/terra-utils/lib/index.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export const normalizeBlock = (data: any): Block => ({
+  hash: data.block_id.hash,
+  timestamp: convertDateToTimestamp(data.block.header.time),
+  size: Number(data.block.header.height),
+  number: Number(data.block.header.height),
+  parentHash: data.block.last_commit.block_id.hash
+})
+
+export const normalizeTransaction = (data: any, asset: string): Transaction<terra.InputTransaction> => {
```
kraikov: I could not the number of confirmations, can you include it please

packages/terra-wallet-provider/lib/TerraWalletProvider.ts
```diff
@@ -0,0 +1,154 @@
...
+  async sendSweepTransaction(address: string | Address): Promise<Transaction<terra.InputTransaction>> {
+    const addresses = await this.getAddresses()
+
+    const balance = await this.getMethod('getBalance')(addresses)
+
+    const message = this._sendMessage(address, balance)
+
+    const fee = await this.getMethod('_estimateFee')(this._signer.accAddress, [message])
+
+    return await this.sendTransaction({ to: address, value: balance.minus(fee * 2) })
```
kraikov, line 112: what's the reason for fee * 2?

what is 2?

packages/types/lib/terra/index.ts
```diff
@@ -0,0 +1,21 @@
...
+  expiration?: number
+  value?: number
+  secret_hash?: string
+  secret?: string
+  contractAddress?: string
+  method?: {
+    claim?: {
+      secret: string
+    }
+    refund: () => void
```
kraikov: is this supposed to be a function?

test/integration/common.ts
```diff
@@ -316,6 +335,25 @@ async function fundAddress(chain: Chain, address: string, value?: BigNumber): Pr
...
+      terra
+        .addProvider(new TerraRpcProvider(config.terra.network))
+        .addProvider(
+          new TerraWalletProvider({
+            network: config.terra.network,
+            mnemonic: config.terra.senderMnemonic,
+            derivationPath: ''
+          })
+        )
+        .addProvider(new TerraSwapProvider())
```
kraikov: No need for the swap provider for send tx.

kraikov (changes requested):

packages/terra-wallet-provider/lib/TerraWalletProvider.ts
AndonMitev: Yes so, _estimateFee always give a bit less than real fee, for example if transaction fee would cost 11000uluna, _estemateFee will return 10000 and transaction are always failing

packages/types/lib/terra/index.ts
AndonMitev: Yes, so by default claim and refund both are functions which are passed to the contract, I've used function for type because `refund: object` or `refund: {}` is not liked by the linter and in js/ts function are objects as well, that why I'm using function, even more I think should convert claim to function as well :)

packages/terra-utils/lib/index.ts
AndonMitev: Yes I'm thinking about them as well because they are needed in the wallet but the transaction it self do not have confirmations prop, [example tx](https://finder.terra.money/tequila-0004/tx/7A58E4C26FC3A3E9883942D89380CE0143275088BE111849BDE8638C34C80EC6), maybe we can depend on status prop or i'll need to count block which are passed from broadcasting tx and if they are `gte(10)` to set specific number of confirmations

kraikov: `confirmations = currentBlock - txBlock`

PR #451, opened 2021-08-20
Feat: Terra upgrade for the new mainnet

AndonMitev:
:warning: THIS IS A PULL REQUEST TEMPLATE! :warning: This helps you complete a pull request. Please follow the instructions below. Reminder that you are bound to abide by the [Code of Conduct](./CODE_OF_CONDUCT.md)

### Description

Describe the purpose of this PR

- [ ] Add x
- [ ] Fix y
- [ ] Check z


test/integration/swap/singleChain.test.ts
```diff
@@ -34,7 +34,7 @@ chai.use(chaiAsPromised)
 const mockSecret = _.repeat('ff', 32)
 
 function testSwap(chain: Chain) {
-  it('Generated secrets are different', async () => {
+  xit('Generated secrets are different', async () => {
```
monokh: What's going on here? Secrets should be unique

packages/terra-rpc-provider/lib/TerraRpcProvider.ts
```diff
@@ -0,0 +1,132 @@
...
+    return promiseBalances
+      .map((balance) => new BigNumber(balance))
+      .reduce((acc, balance) => acc.plus(balance), new BigNumber(0))
+  }
+
+  sendRawTransaction(): Promise<string> {
+    throw new Error('Method not implemented.')
+  }
+
+  async _getTransactionsForAddress(address: Address | string): Promise<any[]> {
```
monokh: Why is this here if only `TerraSwapFindProvider` uses it?
Should move this over there. And make sure to add a type for the response, so we can benefit from the typescript type checking

packages/terra-swap-provider/lib/TerraSwapProvider.ts
```diff
@@ -0,0 +1,125 @@
...
+export default class TerraSwapProvider extends Provider implements Partial<SwapProvider> {
+  private _network: TerraNetwork
+
+  constructor(network: TerraNetwork) {
+    super()
+    this._network = network
+  }
+
+  async generateSecret(message: string): Promise<string> {
+    return sha256(message)
```
monokh: Needs to involve signing message so that it creates a unique secret for different wallets. You don't want someone to be able to generate your secret.

packages/terra-swap-provider/lib/TerraSwapProvider.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    if (!initTx) {
+      throw new TxNotFoundError(`Transaction not found: ${initiationTxHash}`)
+    }
+
+    if (!doesTransactionMatchInitiation(swapParams, initTx._raw)) {
+      throw new StandardError('Transactions are not matching')
+    }
+
+    return true
```
monokh, line 98: Is it possible for the transaction to exist but have a "failed" state? If so should check that

packages/terra-utils/lib/index.ts
```diff
@@ -0,0 +1,94 @@
...
+export const normalizeBlock = (data: any): Block => ({
+  hash: data.block_id.hash,
+  timestamp: convertDateToTimestamp(data.block.header.time),
+  size: Number(data.block.header.height),
+  number: Number(data.block.header.height),
+  parentHash: data.block.last_commit.block_id.hash
+})
+
+export const normalizeTransaction = (
+  data: any,
```
monokh, line 15: why any? Don't all terra transactions have a specific format?

packages/terra-utils/lib/index.ts
```diff
@@ -0,0 +1,94 @@
...
+    logs?.events?.find((e: any) => e.type === 'wasm')?.attributes.find((e: any) => e.key === 'contract_address')
+      .value ||
+    ''
+
+  return {
+    value: Number(value),
+    hash: data.txhash,
+    confirmations: Math.min(currentBlock - data.height, 10),
+    ...(txParams?.secret && { secret: txParams.secret }),
+    _raw: {
```
monokh, line 61: The _raw part of the transaction is supposed to be the actual raw response from the RPC
I think if you want to parse it to get other details, you should write a separate function that does that

monokh (changes requested):

packages/terra-networks/lib/index.ts
```diff
@@ -0,0 +1,44 @@
...
+  name: 'mainnet',
+  networkId: 'mainnet',
+  nodeUrl: 'https://lcd.terra.dev',
+  helperUrl: 'https://fcd.terra.dev/v1',
+  coinType: '397',
+  isTestnet: false,
+  chainID: 'columbus-5',
+  asset: 'luna',
+  codeId: 6431, // TODO: Replace after deploying contract on mainnet
+  gasPrices: 0.01133 // // TODO: Replace after deploying contract on mainnet
```
monokh: This isn't dependent on the network or anything?

packages/terra-rpc-provider/lib/TerraRpcProvider.ts
```diff
@@ -0,0 +1,132 @@
...
+  constructor(network: TerraNetwork) {
+    super({
+      baseURL: network.helperUrl,
+      responseType: 'text',
+      transformResponse: undefined
+    })
+    this._lcdClient = new LCDClient({
+      URL: network.nodeUrl,
+      chainID: network.chainID,
+      gasPrices: network.gasPrices + network.asset
```
monokh: So the fee is always static?

test/integration/swap/singleChain.test.ts
AndonMitev: Fixed

packages/terra-rpc-provider/lib/TerraRpcProvider.ts
AndonMitev: Fixed

packages/terra-swap-provider/lib/TerraSwapProvider.ts
AndonMitev: Fixed

packages/terra-utils/lib/index.ts
AndonMitev: Done

packages/terra-networks/lib/index.ts
AndonMitev: Picking them dynamically from the terra chain

packages/terra-rpc-provider/lib/TerraRpcProvider.ts
AndonMitev: Updated to be dynamic

packages/terra-utils/lib/index.ts
AndonMitev: Where complicated one 🗡️  Yes they do have TxInfo but then most of the values becomes red like they are not recognized/not part of the object which is kinda weird because I extract the from it

packages/terra-swap-provider/lib/TerraSwapProvider.ts
AndonMitev: Will check if they have any failed state

packages/terra-wallet-provider/lib/TerraWalletProvider.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    return signed.toString('hex')
+  }
+
+  async getConnectedNetwork(): Promise<TerraNetwork> {
+    return this._network
+  }
+
+  async sendTransaction(sendOptions: terra.TerraSendOptions): Promise<Transaction<terra.InputTransaction>> {
+    const { to, value, messages } = sendOptions
```
monokh: So the fee parameter is ignored? `options.fee`

Does terra have a concept of lower or higher fees?

AndonMitev: I'll take a look right away

AndonMitev: Added

packages/terra-wallet-provider/lib/TerraWalletProvider.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    const data = sendOptions.data as any
+
+    const msgs = []
+
+    if (data) {
+      return this._handleInjection(data)
```
monokh: Usually we shouldn't refer to some functionality like injection here

What i would do is:
1. Do sendOptions.data as `CreateAndSignTxInput` (whatever the input type to `createAndSignTx is)
2. In this same function spread the sendOptions.data into `createAndSignTx`
3. Only if `to` and `value` are good values, add the `_sendMessage`

I think this way, dont need another function and also `gasPrices` will be defaulted if not exist

monokh (changes requested):

PR #453, opened 2021-09-09
feat: improve tranassctions confirmation speed on eth

tima-t:


monokh:
I think you need to update some tests @tima-t

kraikov:
CAL migrated to Chainify

PR #462, opened 2021-10-11
feat: export private key from js wallets

b0b4456:
https://linear.app/liquality/issue/LIQ-329/implement-getprivatekey-functionality-in-chainabstractionlayer

### DIFF
--- a/packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.ts
+++ b/packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.ts
@@ -57,6 +57,10 @@ export default class BitcoinJsWalletProvider extends BitcoinWalletProvider(
     return ECPair.fromWIF(wif, this._network)
   }
 
+  async getPrivateKey() {
+    return 'FIXME'
+  }
+
   async signMessage(message: string, from: string) {
     const address = await this.getWalletAddress(from)
     const keyPair = await this.keyPair(address.derivationPath)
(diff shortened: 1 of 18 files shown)


packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.ts
```diff
@@ -57,6 +57,11 @@ export default class BitcoinJsWalletProvider extends BitcoinWalletProvider(
     return ECPair.fromWIF(wif, this._network)
   }
 
+  async exportPrivateKey() {
+    const pair = await this.keyPair(this._baseDerivationPath)
+    return pair.privateKey?.toString('hex')
```
harshjv: do we need `?` (can `pair.privateKey` be empty?)

test/integration/wallet/bitcoinWallet.test.ts
```diff
@@ -159,6 +159,13 @@ function testWallet(chain: Chain) {
       expect(signedMessage1).to.equal(signedMessage2)
     })
   })
+
+  describe('exportPrivateKey', () => {
+    it('should return hex string', async () => {
+      const key = await chain.client.wallet.exportPrivateKey()
+      expect(key).not.to.be.empty
```
harshjv: can you update your tests to verify actual priv key instead of `not.to.be.empty`?

harshjv (changes requested):

packages/bitcoin-js-wallet-provider/lib/BitcoinJsWalletProvider.ts
b0b4456: according to typescript it can be empty

test/integration/wallet/bitcoinWallet.test.ts
b0b4456: for eth and near yes, I was able to print out a string in the console and could paste that in. But for btc i can't actually run the tests, is it because I'd need bitcoind running locally ?

harshjv (approved):
LGTM

PR #466, opened 2021-10-21
fix: ETH privateKey format

b0b4456:
before: `exportPrivateKey` would return xpriv/xpub json for ETH
after: `exportPrivateKey` returns 64-character privateKey string that I confirmed is importable into metamask.


b0b4456:
see https://linear.app/liquality/issue/LIQ-391/poc-render-key-in-wallet

monokh (approved):

PR #472, opened 2021-11-01
feat: Handle cw20 terra

AndonMitev:
LIQ-297

https://linear.app/liquality/issue/LIQ-297/need-to-add-terra-chain-option-to-add-custom-token-under-manage-assets


packages/terra-networks/lib/index.ts
```diff
@@ -8,6 +8,7 @@ export interface TerraNetwork extends Network {
   chainID: string
   asset: string
   codeId: number
+  tokenAddress?: string
```
monokh: This feels like the wrong way of constructing this.. I mean the token address is not described by a network
To be honest, i think it should be same for asset, These should rather be constructor parameters to the providers

packages/terra-wallet-provider/lib/TerraWalletProvider.ts
```diff
@@ -29,9 +30,11 @@ export default class TerraWalletProvider extends WalletProvider {
   private _signer: MnemonicKey
   private _lcdClient: LCDClient
   private _wallet: Wallet
+  private _asset: string
+  private _tokenAddress: string
   _accAddressKey: string
 
-  constructor(options: TerraWalletProviderOptions) {
+  constructor(options: TerraWalletProviderOptions, asset: string, tokenAddress?: string) {
```
monokh: `asset`, `tokenAddress` should be in options

monokh: also some check failure hmm

packages/terra-wallet-provider/lib/TerraWalletProvider.ts
```diff
@@ -159,16 +174,18 @@ export default class TerraWalletProvider extends WalletProvider {
       txData = {
         ...(fee && {
           gasPrices: new Coins({
-            [this._network.asset]: fee
+            [this._asset]: fee
           })
         })
       }
     } else {
+      const feeAsset = this._tokenAddress ? 'uluna' : this._asset
```
monokh: I thought that you can pay fee in any asset? Why not just respect whatever the user passed in constructor as `asset` ? 

If it's only possible to pay fee in luna, throw an error if `tokenAddress` is specified and asset is not luna. This avoids situation where a client passes an asset, and they think its going to be used

monokh (changes requested):

monokh (approved):

PR #475, opened 2021-11-09
Fix/calculate max ust fee

AndonMitev:


packages/terra-rpc-provider/lib/TerraRpcProvider.ts
```diff
@@ -135,4 +135,14 @@ export default class TerraRpcProvider extends NodeProvider implements FeeProvide
       }
     }
   }
+
+  async getTaxFees(amount: number, denom = 'uusd'): Promise<any> {
```
monokh: Denom shouldn't be optional I think, It should be explicit which taxes you want to retrieve

packages/terra-rpc-provider/lib/TerraRpcProvider.ts
```diff
@@ -135,4 +135,14 @@ export default class TerraRpcProvider extends NodeProvider implements FeeProvide
...
+    const taxRate = await this._lcdClient.treasury.taxRate()
+    const taxCap = await this._lcdClient.treasury.taxCap(denom)
+
+    const _taxRate = taxRate.toNumber()
+    const _taxCap = taxCap.amount.toNumber()
+
+    return Math.min((amount || 0) * _taxRate, _taxCap / 1_000_000)
```
monokh: Shouldn't amount be always provided? Why the default to 0?

monokh (changes requested):

packages/terra-rpc-provider/lib/TerraRpcProvider.ts
AndonMitev: Because initially Swap.vue pass undefined and evaluates to NaN 😄

monokh (approved):

PR #478, opened 2021-11-30
fix: Handle terra tx status

AndonMitev:
Handle transaction after broadcast on init, if there is any error transaction will be reverted and htlc wont be initialized. Add error handler before claim and refund transaction to check if status of init transaction is success or failure. Set fee when user init swap

### DIFF
--- a/packages/terra-swap-provider/lib/TerraSwapProvider.ts
+++ b/packages/terra-swap-provider/lib/TerraSwapProvider.ts
@@ -93,14 +93,16 @@ export default class TerraSwapProvider extends Provider implements Partial<SwapP
     if (!initTx) {
       throw new TxNotFoundError(`Transaction not found: ${initiationTxHash}`)
     }
-
-    if (!doesTransactionMatchInitiation(swapParams, initTx._raw)) {
-      throw new StandardError('Transactions are not matching')
-    }
-
+    
     if (isTxError(initTx)) {
       throw new StandardError(`Encountered an error while running the transaction: ${initTx.code} ${initTx.codespace}`)
     }
+
+    if (!doesTransactionMatchInitiation(swapParams, initTx['_raw'])) {
+      throw new StandardError('Transactions are not matching')
+    }
+
+    
 
     return true
   }
--- a/packages/terra-swap-provider/lib/TerraSwapProvider.ts
+++ b/packages/terra-swap-provider/lib/TerraSwapProvider.ts
@@ -26,7 +26,7 @@ export default class TerraSwapProvider extends Provider implements Partial<SwapP
     return transaction?.secret
   }
 
-  async initiateSwap(swapParams: SwapParams, fee: any): Promise<Transaction<terra.InputTransaction>> {
+  async initiateSwap(swapParams: SwapParams, fee: number): Promise<Transaction<terra.InputTransaction>> {
     validateSwapParams(swapParams)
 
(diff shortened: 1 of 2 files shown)


kraikov:
LGTM.

@harshjv can you also take a look ?

packages/terra-swap-provider/lib/TerraSwapProvider.ts
```diff
@@ -97,6 +98,10 @@ export default class TerraSwapProvider extends Provider implements Partial<SwapP
       throw new StandardError('Transactions are not matching')
     }
 
+    if (isTxError(initTx)) {
```
harshjv, line 97: Can we check for `isTxError` before `doesTransactionMatchInitiation`?

packages/terra-wallet-provider/lib/TerraWalletProvider.ts
```diff
@@ -198,11 +199,15 @@ export default class TerraWalletProvider extends WalletProvider {
         fee: isProto ? Fee.fromData(JSON.parse(data.fee as any)) : Fee.fromAmino(JSON.parse(data.fee as any))
       }
     } else if (data?.msgs) {
+      const gasPrice = data.fee as any
+      const gasLimit = 800000
```
harshjv, line 203: Do we want a static `gasLimit` in Terra?

harshjv (changes requested):

packages/terra-wallet-provider/lib/TerraWalletProvider.ts
AndonMitev: Yeah we do and this is used only for init operation. The reason why we need it is to control the spended amount for fee on Init Swap transaction and to match perfectly with the wallet otherwise they are setting their own and sometimes fee could be different in the explorer than the one displayed in the wallet, here is a related task to this issue: https://linear.app/liquality/issue/LIQ-649/calculate-the-max-fee-so-that-there-is-gas-left-for-the-user-to-pay

kraikov: The caller can set the desired fee by passing `data.fee` (checked in the above `if` statement). That's the way most Terra dApps are working and requesting.

The static `gasLimit` is only used for `initiateSwap/claim/refund`

The same amount is used in the wallet. It's more than enough to cover the required gas limit for `initiateSwap/claim/refund`

packages/terra-swap-provider/lib/TerraSwapProvider.ts
AndonMitev: done

kraikov: Do we want to make external calls/http calls before simple boolean checks? It's suboptimal to make external calls before checking if the params are matching IMO.

AndonMitev: @harshjv wdyt and can we merge the code and update agent && wallet or there is something else which needs to be fixed

AndonMitev: Hello any updates on this @harshjv

harshjv: @kraikov `isTxError` is a synchronous method, so i do not think that its making any HTTP request.

harshjv (approved):

PR #479, opened 2021-11-30
fix: check if status has SuccessValue

kraikov:
Successful tx:
Screenshot 2021-11-30 at 21 06 18

Failed tx:
Screenshot 2021-11-30 at 21 06 54


beemi (approved):
LGTM

PR #480, opened 2021-12-01
fix: Check transaction status for solana and throw error if is not successful

AndonMitev:
Add handler to check status of the transactions for solana and throw error if they are not successful


tima-t:
LGTM

tima-t (approved):

harshjv (approved):

PR #481, opened 2021-12-02
Fix: Compare transaction lamports againts swapParams value

AndonMitev:
Check if transaction lamports are equal to swap params.value when we are comparing transactions. That way we are using the real provided lamport value when user is creating a htlc and not the passed one in init params


kraikov (approved):
LGTM.

harshjv (approved):

PR #490, opened 2021-12-22
Add support for EIP 1599

harshjv:


packages/bitcoin-node-wallet-provider/lib/BitcoinNodeWalletProvider.ts
```diff
@@ -73,7 +73,7 @@ export default class BitcoinNodeWalletProvider extends WalletProvider {
 
   async sendTransaction(options: SendOptions) {
     return options.fee
-      ? this.withTxFee(async () => this._sendTransaction(options), options.fee)
+      ? this.withTxFee(async () => this._sendTransaction(options), options.fee as number)
```
monokh, line 76: Hmm weird, why do we need to cast `as number` everywhere? It looks like the places you are passing this, the type is already `number` and the destination type is `number`

packages/ethereum-eip1559-fee-provider/lib/EthereumEIP1559FeeProvider.ts
```diff
@@ -0,0 +1,188 @@
...
+  _slowMultiplier: number
+  _averageMultiplier: number
+  _fastMultiplier: number
+  _maxFeePerGasThreshold: number
+
+  constructor(options: { uri: string; username?: string; password?: string }, opts: FeeOptions = {}) {
+    super(options.uri, options.username, options.password)
+
+    const { slowMultiplier = 1, averageMultiplier = 1.25, fastMultiplier = 1.4, maxFeePerGasThreshold = 1000 } = opts
+    this._slowMultiplier = slowMultiplier
```
monokh: are the multipliers needed anymore?
I mean i thought the EIP1559 logic itself can give good recommendations
the main use case of this multiplier was that the estimation come from node for legacy transaction was not very accurate

packages/ethereum-ledger-provider/lib/EthereumLedgerProvider.ts
```diff
@@ -68,47 +72,87 @@ export default class EthereumLedgerProvider extends LedgerProvider<HwAppEthereum
...
     const txOptions: ethereum.UnsignedTransaction = {
       from,
       to: options.to ? addressToString(options.to) : (options.to as string),
       value: options.value,
       data: options.data,
-      gasPrice,
       nonce
     }
 
+    if (options.fee) {
```
monokh: Wow there is a lot of duplication with JsProvider 
Definitely will be good to have a EthereumWalletProvider base class
We can do separately/later

monokh (approved):

packages/ethereum-eip1559-fee-provider/lib/EthereumEIP1559FeeProvider.ts
harshjv: Started using https://github.com/rainbow-me/fee-suggestions and it seems to be handling this well with surge in fees.

PR #493, opened 2022-01-12
fix/near claim expiration check

kraikov:
### Description

- Remove expiration check when claiming a swap.

- The bytecode is taken from [the near contract](https://github.com/weichain/near-htlc-contracts) - In order to validate the change, you can clone the repo, install the dependencies in both the root and contract forlders and run **npm build.** 
This will generate a bytecode.json file that includes the bytecode generated after compiling the contract.

- The integration tests now check that swaps are claimable after the expiration
Screenshot 2022-01-12 at 12 16 18


test/integration/swap/singleChain.test.ts
```diff
@@ -425,6 +451,14 @@ function testFee(chain: Chain) {
 describe('Swap Single Chain Flow', function () {
   this.timeout(TEST_TIMEOUT)
 
+  describeExternal('Claim works after expiration', () => {
```
monokh: Why is this in a separate section like this? Instead of another test in `testSwap`? 
This test should pass for every swap situation

monokh (changes requested):

test/integration/swap/singleChain.test.ts
kraikov: This could cause test failing sometimes due to the nature of how tests are done for Near, Solana and Terra (testnets not local nodes)

monokh: in that case `testClaimSucceedAfterExpiration` should be called in each appropriate section
like `Bitcoin - Node`

This way, it's properly covered every scenario but still can pick an choose which chain to test

monokh (approved):

harshjv (approved):

packages/near-swap-provider/lib/NearSwapProvider.ts
```diff
@@ -13,7 +13,7 @@ import {
 
 import Bytecode from './bytecode'
 
-const CONTRACT_CODE = 'wnGY4a+YYkfbkBqYqyhdQjjtHRl9Auyh7yxCGtXL8JI='
+const CONTRACT_CODE = 'jrBWhtpuyGJ44vtP+Ib+I32tuUUtfKQBuBdQ8y3M6Ro='
```
harshjv, line 16: @kraikov how can one audit/review this contract code?

kraikov: Not sure how this is generated. Probably it's baseXX of the hash of the bytecode. However, once you verify the bytecode is good, you can create swap transaction, read the metadata of the transaction from Near and there you'll find transaction.code

PR #494, opened 2022-01-12
NFT support

skeremidchiev:
### Description

This PR adds support for NFT (ERC721 & ERC1155) to CAL.
It relies on [OpenSeaAPI](https://docs.opensea.io) to fetch all user nfts in a single call.


packages/nft-erc721-provider/lib/NftErc721Provider.ts
```diff
@@ -0,0 +1,62 @@
...
+  constructor(options: { uri: string; mnemonic: string; derivationPath: string }) {
+    super()
+
+    this._wallet = Wallet.fromMnemonic(options.mnemonic, options.derivationPath)
+    this._wallet = this._wallet.connect(new JsonRpcProvider(options.uri))
+    this._contract = new Contract('0x0000000000000000000000000000000000000000', NftErc721_ABI, this._wallet)
+  }
+
+  private _attach(contract: Address | string) {
+    if (this._contract.address != contract) this._contract = this._contract.attach(ensure0x(addressToString(contract)))
```
kraikov: Use parentheses even for single line statements please

packages/nft-erc721-provider/lib/NftErc721Provider.ts
```diff
@@ -0,0 +1,62 @@
...
+  async approve(contract: Address | string, receiver: Address | string, tokenId: BigNumber) {
+    this._attach(contract)
+
+    const txWithHash = await this._contract.functions.approve(ensure0x(addressToString(receiver)), tokenId.toString())
+    return normalizeTransactionObject(txWithHash)
+  }
+
+  async approveAll(contract: Address | string, receiver: Address | string, state?: boolean) {
+    this._attach(contract)
+    if (state === undefined) state = true
```
kraikov: Use default parameter for state please.
https://www.typescripttutorial.net/typescript-tutorial/typescript-default-parameters/

packages/nft-erc721-provider/lib/NftErc721Provider.ts
```diff
@@ -0,0 +1,62 @@
...
+export default class NftErc721Provider extends Provider implements Partial<NftProvider> {
+  _contract: Contract
+  _wallet: Wallet
+  _jsonRpcProvider: JsonRpcProvider
+
+  constructor(options: { uri: string; mnemonic: string; derivationPath: string }) {
+    super()
+
+    this._wallet = Wallet.fromMnemonic(options.mnemonic, options.derivationPath)
+    this._wallet = this._wallet.connect(new JsonRpcProvider(options.uri))
```
kraikov: An ethers Provider will execute frequent getNetwork calls to ensure the network calls and network being communicated with are consistent.
In our case this is not the case, so you can use `StaticJsonRpcProvider` instead.

packages/nft-erc721-provider/lib/NftErc721Provider.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export default class NftErc721Provider extends Provider implements Partial<NftProvider> {
+  _contract: Contract
+  _wallet: Wallet
+  _jsonRpcProvider: JsonRpcProvider
+
+  constructor(options: { uri: string; mnemonic: string; derivationPath: string }) {
+    super()
+
+    this._wallet = Wallet.fromMnemonic(options.mnemonic, options.derivationPath)
```
kraikov: Is it possible to pass the provider here instead of using `connect` later?

kraikov (changes requested):

packages/nft-erc721-provider/lib/NftErc721Provider.ts
```diff
@@ -0,0 +1,62 @@
...
+import { ensure0x, normalizeTransactionObject } from '@liquality/ethereum-utils'
+import { NftProvider, Address, BigNumber } from '@liquality/types'
+
+import { Contract } from '@ethersproject/contracts'
+import { JsonRpcProvider } from '@ethersproject/providers'
+import { Wallet } from '@ethersproject/wallet'
+
+import NftErc721_ABI from './NftErc721_ABI.json'
+
+export default class NftErc721Provider extends Provider implements Partial<NftProvider> {
```
kraikov: Why the `NftErc721Provider` implements Partially the `NftProvider`? What's missing?

kraikov:
Overall looks good to me so far. How do you plan to manage the differences between Erc721 and Erc1155?

packages/nft-erc721-provider/lib/NftErc721Provider.ts
skeremidchiev: No:

static fromMnemonic(mnemonic: string, path?: string, wordlist?: Wordlist): Wallet {
    if (!path) { path = defaultPath; }
    return new Wallet(HDNode.fromMnemonic(mnemonic, null, wordlist).derivePath(path));
}

skeremidchiev:
It seems like the Nft interface will have to be extended a bit to match the ERC1151 functionality.

packages/nft-erc1155-provider/lib/NftErc1155Provider.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    if (tokenID) {
+      const amount = await this._contract.functions.balanceOf(await this._wallet.getAddress(), tokenID)
+      return amount[0].toNumber()
+    }
+
+    return 0
+  }
+
+  async transfer(
```
kraikov, line 40: I believe you have to check wether the `tokenIDs` is an array and if it's an array you should use `safeBatchTransferFrom` otherwise you should use `safeTransferFrom`

packages/nft-erc1155-provider/lib/NftErc1155Provider.ts
```diff
@@ -0,0 +1,80 @@
...
+    this._contract = new Contract('0x0000000000000000000000000000000000000000', NftErc1155_ABI, this._wallet)
+  }
+
+  private _attach(contract: Address | string) {
+    if (this._contract.address !== contract) {
+      this._contract = this._contract.attach(ensure0x(addressToString(contract)))
+    }
+  }
+
+  async balance(contract: Address | string, tokenID?: number) {
```
kraikov: This function should also cover `balanceOfBatch`

packages/nft-erc1155-provider/lib/NftErc1155Provider.ts
```diff
@@ -0,0 +1,80 @@
...
+  ) {
+    this._attach(contract)
+
+    if (tokenIDs && tokenIDs.constructor === Array) {
+      const txWithHash = await this._contract.functions.safeBatchTransferFrom(
+        await this._wallet.getAddress(),
+        ensure0x(addressToString(receiver)),
+        tokenIDs,
+        values,
+        '0x00'
```
kraikov: SInce we're building a library we should allow passing of that parameter, because we don't know how the consumer will use it. It could have default value of '0x00' but it should be possible to pass other values.

kraikov (changes requested):

packages/nft-erc721-provider/lib/NftErc721Provider.ts
skeremidchiev: It's WIP. But I'm thinking to add the fetch NFTs method to NftProvider Interface.

packages/nft-erc1155-provider/lib/NftErc1155Provider.ts
skeremidchiev: Nice point.

packages/nft-erc1155-provider/lib/NftErc1155Provider.ts
skeremidchiev: Looking into it.

packages/nft-erc1155-provider/lib/NftErc1155Provider.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    this._signer = signer
+    this._contract = new Contract('0x0000000000000000000000000000000000000000', NftErc1155_ABI, this._signer)
+    this._contractCache = {}
+  }
+
+  async balance(contract: Address | string, tokenIDs: number | number[]) {
+    await this._attach(contract)
+
+    if (!tokenIDs) return 0
```
kraikov: Use { } please.

packages/nft-erc1155-provider/lib/NftErc1155Provider.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    if (tokenIDs.constructor === Array) {
```
kraikov: Use 
`import { isArray } from 'lodash'` instead

packages/nft-erc1155-provider/lib/NftErc1155Provider.ts
```diff
@@ -0,0 +1,126 @@
...
+      const addresses = Array.from({ length: tokenIDs.length }).fill(await this._signer.getAddress())
```
kraikov: You should do `await this._signer.getAddress()` only once outside the fill method.

packages/nft-erc1155-provider/lib/NftErc1155Provider.ts
```diff
@@ -0,0 +1,126 @@
...
+    receiver: Address | string,
+    tokenIDs: number | number[],
+    values: number[],
+    data = '0x00'
+  ) {
+    await this._attach(contract)
+
+    if (
+      tokenIDs &&
+      tokenIDs.constructor === Array &&
```
kraikov: Use `import { isArray } from 'lodash'` here

packages/nft-erc1155-provider/lib/NftErc1155Provider.ts
```diff
@@ -0,0 +1,126 @@
...
+      values &&
+      values.constructor === Array &&
```
kraikov: Use `import { isArray } from 'lodash'` here

packages/nft-erc1155-provider/lib/NftErc1155Provider.ts
```diff
@@ -0,0 +1,126 @@
...
+    tokenIDs: number | number[],
+    values: number[],
+    data = '0x00'
+  ) {
+    await this._attach(contract)
+
+    if (
+      tokenIDs &&
+      tokenIDs.constructor === Array &&
+      values &&
```
kraikov: You can combine the checks like this:
`isArray(values?.constructor)`

so `values && values.constructor === Array` becomes ``isArray(values?.constructor)``

packages/nft-erc1155-provider/lib/NftErc1155Provider.ts
```diff
@@ -0,0 +1,126 @@
...
+      values.constructor === Array &&
+      tokenIDs.length === values.length
+    ) {
+      let txWithHash
+      if (tokenIDs.length === 1) {
```
kraikov: What happens if `tokenIDs` is not an array?

packages/nft-erc721-provider/lib/NftErc721Provider.ts
```diff
@@ -0,0 +1,103 @@
...
+      await this._signer.getAddress(),
+      ensure0x(addressToString(operator))
+    )
+    return state[0]
+  }
+
+  private async _supportsInterface(contractInstance: Contract) {
+    const contractAddress = contractInstance.address.toLowerCase()
+
+    if (this._contractCache[contractAddress] !== undefined) {
```
kraikov: Shall we replace that with
`if (this._contractCache[contractAddress])`

kraikov (changes requested):

packages/nft-erc1155-provider/lib/NftErc1155Provider.ts
skeremidchiev: In this case `transfer` method `throw`s an error.

packages/nft-erc721-provider/lib/NftErc721Provider.ts
skeremidchiev: I don't think so because this map: `_contractCache: { [address: string]: boolean }` contains boolean values.

packages/nft-erc1155-provider/lib/NftErc1155Provider.ts
```diff
@@ -0,0 +1,87 @@
...
+  constructor(signer: Signer) {
+    super(signer, NftErc1155_ABI as any, erc1155InterfaceID)
+  }
+
+  async balance(contract: Address | string, tokenIDs: number | number[]) {
+    if (!tokenIDs) {
+      return 0
+    }
+
+    await super.balance(contract, tokenIDs)
```
monokh: From here it's not clear what the super function is doing. If all the super functions are just attaching the contract and caching it, then I think its better to call it something else.

like `this.setContract(contract)`

as a result I think the super methods are unnecessary

packages/client/lib/Nft.ts
```diff
@@ -0,0 +1,60 @@
...
+  async balance(contract: Address | string, tokenIDs?: number | number[]): Promise<number | number[]> {
+    const balance = await this.client.getMethod('balance')(contract, tokenIDs)
+    return balance
+  }
+
+  /** @inheritdoc */
+  async transfer(
+    contract: Address | string,
+    receiver: Address | string,
+    tokenIDs: number | number[],
```
monokh: Do most nft standards have ability to transfer multiple in 1 call? If not, best to have to separate functions. `transfer` and `transferBatch`

This way, a provider could implement either, or throw an error if that function is not possible

packages/nft-erc1155-provider/lib/NftErc1155Provider.ts
```diff
@@ -0,0 +1,87 @@
...
+  async transfer(
+    contract: Address | string,
+    receiver: Address | string,
+    tokenIDs: number[],
+    values: number[],
+    data = '0x00'
+  ) {
+    await super.transfer(contract, receiver, tokenIDs, values, data)
+
+    if (isArray(tokenIDs) && isArray(values) && tokenIDs.length === values.length) {
```
monokh: Better to get the errors out of the way at the start of the function rather than end
And I think you should throw an appropriate error for each scenario. I.e. it is not array and lengths dont match

packages/nft-erc721-provider/lib/NftErc721Provider.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  async balance(contract: Address | string) {
+    await super.balance(contract)
+
+    const amount = await this._contract.functions.balanceOf(await this._signer.getAddress())
+    return amount[0].toNumber()
+  }
+
+  async transfer(contract: Address | string, receiver: Address | string, tokenID: number) {
```
monokh: Since this nft standard does not let you transfer batch, I think its important to make `transfer` for single and anohter method `transferBatch` for multiple. 

A user may add this provider and then call `client.nft.transfer` with multiple tokenIDs and expect that it should work

packages/nft-provider/lib/NftProvider.ts
```diff
@@ -0,0 +1,97 @@
...
+    super({
+      baseURL: apiURI,
+      responseType: 'text',
+      transformResponse: undefined,
+      headers: {
+        'X-Api-Key': apiKey
+      }
+    })
+
+    this._wallet = Wallet.fromMnemonic(options.mnemonic, options.derivationPath)
```
monokh: This kind of architecture is going to be problematic if we were to support other types of wallets for example ledger... it would require a provider for every wallet type?

when wanting access to the wallet, it should work to use the `this.getMethod('sendTransaction')` method? So instead of attaching the ethersjs wallet and calling the contract, we could serielise the data needed for the call and use `sendTransaction` to make it happen. This way all types of wallets will be supported. I think in ethers its like `contract.transfer.populateTransaction(...)`

In general, this provider is really not standard for the architecture of CAL

monokh (changes requested):

packages/nft-erc721-provider/lib/NftErc721Provider.ts
kraikov: The idea here is that the end user/dev would only use the `NftProvider` without knowing about the different standards (721 & 1155).

packages/nft-erc1155-provider/lib/NftErc1155Provider.ts
kraikov: You would end up with lots of code repetition because `setContract` is the same for both 721 and 1155, so it should probably be in a base provider.

packages/client/lib/Chain.ts
```diff
@@ -101,6 +101,11 @@ export default class Chain implements ChainProvider, FeeProvider {
     return transaction
   }
 
+  /** @inheritdoc */
+  async call(to: Address | string, data: string): Promise<string> {
```
monokh: This wouldn't have meaning in a non ethereum context? So it shouldn't be in the Chain interface, this place is reserved for functions that can be implemented cross chain.
You don't have to have something in the `Chain` to access it from providers. Just use `this.getMethod` in the provider directly

packages/nft-provider/lib/NftProvider.ts
```diff
@@ -0,0 +1,112 @@
...
+    this._subProviders = {}
+    this._subProviders['ERC721'] = new NftErc721Provider()
+    this._subProviders['ERC1155'] = new NftErc1155Provider()
+
+    this._nftContractsCache = {}
+    this._clientAttached = false
+  }
+
+  async balance(contract: Address | string, tokenIDs: number | number[]) {
+    this._setClientToSubProviders()
```
monokh: Why does this happen before each function instread of just once?
I think you can override `setProvider` to do this very simply

packages/client/lib/Nft.ts
monokh: If we are not going to separate to transfer-transferBatch function, we should atleast make this function always take tokenIDs as array and in the nft standard that can't send batch: throw an error when the list contains more than one

monokh (changes requested):

monokh (approved):
LGTM!

kraikov (approved):

Jennievon (approved):
LGTM

PR #497, opened 2022-01-27
fix: Return 0 when balance is undefined and fix tx confirmations

AndonMitev:
This will fix:
[Fix transaction status](https://linear.app/liquality/issue/LIQ-999/terra-send-transactions-was-successful-on-terra-explorer-but-on-wallet)
[NaN when balance is 0](https://linear.app/liquality/issue/LIQ-1003/terra-balance-showing-as-nan-on-overview-screen)

### DIFF
--- a/packages/terra-utils/lib/index.ts
+++ b/packages/terra-utils/lib/index.ts
@@ -57,12 +57,7 @@ export const normalizeTransaction = (
       .value ||
     ''
 
-  let status
-  if (currentBlock - data.height < 10) {
-    status = TxStatus.Pending
-  } else {
-    status = data.raw_log?.includes('failed to execute message') ? TxStatus.Failed : TxStatus.Success
-  }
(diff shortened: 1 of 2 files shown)


packages/terra-utils/lib/index.ts
```diff
@@ -57,12 +57,7 @@ export const normalizeTransaction = (
       .value ||
     ''
 
-  let status
-  if (currentBlock - data.height < 10) {
-    status = TxStatus.Pending
-  } else {
-    status = data.raw_log?.includes('failed to execute message') ? TxStatus.Failed : TxStatus.Success
-  }
+  const status = data.raw_log?.includes('failed to execute message') ? TxStatus.Failed : TxStatus.Success
```
monokh, line 60: is that error message the only way a tx can fail ?

monokh (approved):

PR #503, opened 2022-02-03
feat: fetch multiple addresses

skeremidchiev:
### Description

Fetch NFTs from multiple addresses.


kraikov:
CAL migrated to Chainify

PR #507, opened 2022-02-08
feat: add currency class

skeremidchiev:


monokh:
What's the idea? How does this differ with https://github.com/liquality/cryptoassets ?

kraikov:
CAL migrated to Chainify

PR #509, opened 2022-02-09
feat: Add log error

AndonMitev:
Add log for errors on transactions


kraikov:
CAL migrated to Chainify

PR #529, opened 2022-03-31
fix: Account index

AndonMitev:
https://linear.app/liquality/issue/LIQ-647/multi-account-when-adding-the-multiple-accounts-for-terra-same-account

### DIFF
--- a/packages/terra-wallet-provider/lib/TerraWalletProvider.ts
+++ b/packages/terra-wallet-provider/lib/TerraWalletProvider.ts
@@ -26,6 +26,7 @@ interface TerraWalletProviderOptions {
   feeAsset: string
   tokenAddress?: string
   stableFee?: boolean
+  index?: number
 }
 
 interface CustomTxOptions extends CreateTxOptions {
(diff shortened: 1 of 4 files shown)


packages/terra-wallet-provider/lib/TerraWalletProvider.ts
```diff
@@ -81,12 +84,13 @@ export default class TerraWalletProvider extends WalletProvider {
...
     const wallet = new MnemonicKey({
-      mnemonic: this._mnemonic
+      mnemonic: this._mnemonic,
+      index: this._index
     })
 
     const result = new Address({
       address: wallet.accAddress,
-      derivationPath: this._baseDerivationPath + `/0/0`,
+      derivationPath: this._baseDerivationPath + `/0/${this._index}`,
```
monokh: In the other providers, we have requested for the full path to be passed, instead of separate base + index
I think this is best because it gives flexibility to the client

AndonMitev: Changed applied

monokh (approved):

PR #552, opened 2022-06-07
feat: Integrate jupiter

AndonMitev:


packages/solana/lib/index.ts
```diff
@@ -1,4 +1,5 @@
 export { SolanaChainProvider } from './chain/SolanaChainProvider';
 export { SolanaNetworks } from './networks';
+export * as TerraTypes from './types';
```
kraikov: Solana

kraikov (changes requested):

packages/solana/lib/wallet/SolanaWalletProvider.ts
```diff
@@ -14,10 +14,17 @@ import {
...
+    createAccount,
+    createTransferInstruction,
+    getAccount,
+    getAssociatedTokenAddress,
+    TokenAccountNotFoundError,
+} from '@solana/spl-token';
 import { Connection, Keypair, PublicKey, SystemProgram, Transaction as SolTransaction, TransactionInstruction } from '@solana/web3.js';
 import { mnemonicToSeedSync } from 'bip39';
 import { derivePath } from 'ed25519-hd-key';
+import { SolanaTxRequest } from 'lib/types';
```
kraikov: this will not compile

kraikov (changes requested):

packages/solana/lib/wallet/SolanaWalletProvider.ts
```diff
@@ -83,38 +90,54 @@ export class SolanaWalletProvider extends Wallet<Connection, Promise<Keypair>> {
...
+                const [fromTokenAccount, toTokenAccount] = await Promise.all([
+                    getAssociatedTokenAddress(contractAddress, this._signer.publicKey),
+                    getAssociatedTokenAddress(contractAddress, to),
+                ]);
+
+                try {
+                    await getAccount(this.chainProvider.getProvider(), toTokenAccount);
+                } catch (err) {
+                    if (err instanceof TokenAccountNotFoundError) {
+                        await createAccount(this.chainProvider.getProvider(), this._signer, contractAddress, to);
```
monokh, line 119: Does this send an extra transaction from the user account in certain situations? 
If this is the case, it is happening without the user realising? And that feels wrong

AndonMitev: I agree but thats how the solana in general works, even phantom wallet behave at the same way, they are creating token accounts for you under the hood

AndonMitev: Also transactions have fixed fee which is `00000000.5 SOL` which is very low low amount in cent value, thats why i assume they are doing this under the hood

packages/solana/lib/wallet/SolanaWalletProvider.ts
AndonMitev: resolved

AndonMitev:
@kraikov @monokh i have added NFT provider as well

PR #555, opened 2022-06-13
fix: move NFTAsset type to global level

monokh:
Move the nft asset to global level so that calling `client.nft.fetch` is correctly typed.


kraikov (approved):
LGTM

PR #557, opened 2022-06-13
fix: catch any errors during balance fetching

kraikov:


packages/evm/lib/chain/EvmChainProvider.ts
```diff
@@ -156,16 +156,22 @@ export class EvmChainProvider extends Chain<StaticJsonRpcProvider> {
...
+                    } else {
+                        return ERC20__factory.connect(a.contractAddress, this.provider).balanceOf(user);
+                    }
+                } catch (err) {
+                    this._logger.debug(`no multicall getBalance error ${a}`, err);
+                    return null;
                 }
             });
-            return (await Promise.all(allBalancePromise)).map((b) => new BigNumber(b.toString()));
+            const balances = (await Promise.all(allBalancePromise)).map((b) => (b ? new BigNumber(b.toString()) : new BigNumber(0)));
```
monokh: If i understand correctly, this will return 0 for balances that were not fetched. 
But that doesn't seem right. It should rather return null or exclude from the list. This way the client can differentiate between an actual 0 balance and a missing balance

monokh (changes requested):

packages/evm/lib/chain/EvmChainProvider.ts
kraikov: What is the client going to do with `null`? Set `0` ? Do nothing? 
Excluding from the list is not an option, because then you can't map the asset to the balance.

monokh (approved):

PR #564, opened 2022-06-23
fix: replace toString with toFixed to prevent exponent error

skeremidchiev:
Fix for exponent number issues.

[LIQ-1446](https://linear.app/liquality/issue/LIQ-1446/as-a-user-i-want-to-be-able-to-swap-greater-than-1000-matic)


packages/evm/lib/utils.ts
```diff
@@ -28,7 +28,7 @@ export function toEthereumTxRequest(tx: EthersPopulatedTransaction, fee: FeeType
 
 export function parseSwapParams(tx: SwapParams): ILiqualityHTLC.HTLCDataStruct {
     return {
-        amount: tx.value.toString(),
+        amount: tx.value.toFixed(),
```
kraikov: please use `toString(10)`

PR #565, opened 2022-06-24
feat: add eip1559 fee api provider

kraikov:


harshjv (approved):
LGTM. However, I am not sure if `metaswap.codefi.network` is a public API and meant to be used outside MM.

PR #571, opened 2022-07-06
Fix: Update nft when swap changes & vise-versa

koderholic:
Bug:
When connect method is called with the Swap provider, the wallet gets updated to the wallet instance in the swap provider, but the wallet instance in the NFT provider isn't changed and vice-versa, which can lead to unexpected behaviors.

Fix:
Update the wallet instance in NFT when swap is changed and vise-versa

### DIFF
--- a/packages/client/lib/Client.ts
+++ b/packages/client/lib/Client.ts
@@ -48,6 +48,9 @@ export default class Client<
             case provider instanceof Swap: {
                 this.swap = provider as SwapType;
                 this.connectWallet(this.swap);
+                if (this.nft) {
+                    this.nft.setWallet(this.wallet);
+                }
                 this.connectChain();
                 break;
             }
@@ -55,6 +58,9 @@ export default class Client<
             case provider instanceof Nft: {
                 this._nft = provider as NftType;
                 this.connectWallet(this.nft);
+                if (this.swap) {
+                    this.swap.setWallet(this.wallet);
+                }
                 this.connectChain();
                 break;
             }


packages/client/lib/Client.ts
```diff
@@ -48,13 +48,19 @@ export default class Client<
             case provider instanceof Swap: {
                 this.swap = provider as SwapType;
                 this.connectWallet(this.swap);
+                if (this.nft) {
```
kraikov, line 51: do you believe we can do that in connectWallet method?

packages/client/lib/Client.ts
```diff
@@ -48,13 +48,19 @@ export default class Client<
...
+                    this.nft.setWallet(this.wallet);
+                }
                 this.connectChain();
                 break;
             }
 
             case provider instanceof Nft: {
                 this._nft = provider as NftType;
                 this.connectWallet(this.nft);
+                if (this.swap) {
```
kraikov, line 61: do you believe we can do that in connectWallet method?

kraikov (changes requested):

packages/client/lib/Client.ts
koderholic: Yes, I thought of that earlier, but seems more if statements will be needed to do it in connectWallet, compared to what we have now.  Lemme know what you think.

```
 private connectWallet(source: SwapType | NftType) {
        const wallet = source?.getWallet() as WalletType;
        if (wallet) {
            this.wallet = wallet;
        }
        if (source instanceof Swap) {
            if (this.nft) {
                this.nft.setWallet(this.wallet);
            }
        } else {
            if (this.swap) {
                this.swap.setWallet(this.wallet);
            }
        }
    }
```

kraikov: okay, let's leave it that way for now.

kraikov (approved):

PR #580, opened 2022-08-11
fix: Solana fetching token by using official registry and some leftov…

AndonMitev:
fix: Solana fetching token by using official registry and some leftover code that i forgot to push

### DIFF
--- a/packages/solana/package.json
+++ b/packages/solana/package.json
@@ -21,6 +21,7 @@
         "@chainify/types": "workspace:*",
         "@chainify/utils": "workspace:*",
         "@solana/spl-token": "0.2.0",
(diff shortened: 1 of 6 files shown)


packages/solana/package.json
```diff
@@ -21,6 +21,7 @@
         "@chainify/types": "workspace:*",
         "@chainify/utils": "workspace:*",
         "@solana/spl-token": "0.2.0",
+        "@solana/spl-token-registry": "^0.2.4574",
```
kraikov: pin versions

kraikov (changes requested):

packages/solana/package.json
AndonMitev: Done

PR #612, opened 2022-09-08
feat: add  prefix to IPFS URL

Jennievon:


packages/evm/lib/nft/MoralisNftProvider.ts
```diff
@@ -53,13 +53,16 @@ export class MoralisNftProvider extends EvmNftProvider {
 
                 if (metadata) {
                     const parsed = JSON.parse(metadata);
+                    const image = parsed.image.includes('ipfs://')
+                        ? parsed.image.replace('ipfs://', 'https://ipfs.io/ipfs/')
```
kraikov: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace

replace will not do anything if the pattern is not found so no need to check anything..

kraikov (changes requested):

monokh (commented):
Please add a changeset

